PR #23, opened 2014-04-02
Update DetourNavMeshQuery.cpp

Esenthel:
bug-fix

Thanks for great library


grahamboree:
This is a really old PR with a pretty clear conclusion:
The current change is not acceptable.  Adding asserts to ensure all required parameters are not null at the beginning of this function (and others for that matter) would be a good fix, however.

PR #64, opened 2015-01-01
DetourNode: Change node index from unsigned short to unsigned int.

cexikitin:
On large tiled maps with high precision we could run out of indexs and detour would get stuck in an infinite loop.

Fixes: https://github.com/memononen/recastnavigation/issues/45

### DIFF
--- a/Detour/Include/DetourNode.h
+++ b/Detour/Include/DetourNode.h
@@ -28,7 +28,7 @@ enum dtNodeFlags
 	DT_NODE_PARENT_DETACHED = 0x04, // parent of the node is not adjacent. Found using raycast.
 };
 
-typedef unsigned short dtNodeIndex;
+typedef unsigned int dtNodeIndex;
 static const dtNodeIndex DT_NULL_IDX = (dtNodeIndex)~0;
 
 struct dtNode


hymerman:
Shouldn't this only be an issue if a `dtNodePool` is constructed with `maxNodes` greater than 65535? Perhaps an assert to check for that in the `dtNodePool` would be appropriate.

That seems like rather high number of nodes (the samples use max node counts of 512 and 4096) but if it's legitimately being hit then perhaps this change is necessary. It's just a shame because it bumps up the memory usage of queries somewhat. I can't think of a way to make the index size configurable in a tidy way.

cexikitin:
We are using r&d for serverside navigation of NPCs on very large maps. We were hitting a deadlock as referenced in issue 45 until I sat down and looked at what was happening. I don't think a 2 byte increase will have any considerable impact in the overall consumption of memory

hymerman:
It's (roughly) 2 * `maxNodes` per query, so 128KB if you're talking of node counts around 64k. Not a massive amount, no, but enough to make me pause to think. Bear in mind that Detour is used by projects with tight memory requirements as well, hence all the small data types and bitfields and whatnot.

Just to be clear though, I think this change is fine, I'm just commenting.

Off-topic: how frequently are you doing these searches that visit over 64k nodes? Isn't that quite CPU-intensive? Perhaps a hierarchical approach would help?

cexikitin:
Quite often, and not really the searches get done in ~<5ms.
Also, if your using a large node count there's a good chance you'll get stuck in an infinite loop without this.  If your using small node counts then the size increase doesn't really effect you.

memononen:
One option would be to do a similar treatment as with `DT_POLYREF64`, that is add a define that allows you to bump the index to 32bit.

As @hymerman pointed out, we should add an assert in the `dtNodePool`constructor to check that `maxNodes` is less than `DT_NULL_IDX` and less than `1<<24` (pidx in dtNode is 24 bits).

And alternative would be to just bump it to int (still should add that assert maxNodes < 1<<24, though). Objections?

hymerman:
I think adding a define is a good idea. Perhaps one day there could be a
'dtConfig.h' where this kind of thing lives, which is the place you go to
customise all of the library, from type sizes (appropriate to your use
case) to assert macros, memory hooks, threading and so on.

Asserts are essential for this to work though; the user needs to know when
their config isn't sufficient.

@cexikitin (just understood your username now that I've said it!) - fair
enough! I'd get slaughtered if my pathfinds took 5ms ;) We just have very
different use cases!

jakobbotsch:
The assertion was added in #179. I will close this since the typedef for the node indices in `DetourNode.h` makes it easy to adjust the size for the user's own use case.

PR #65, opened 2015-01-01
DetourAlloc: Use size_t instead of int to better match the amount of memory we could allocate.

cexikitin:
...


grahamboree:
rcAlloc could use this as well.

grahamboree:
@cexikitin This would be a fine change so long as it was consistent with rcAlloc.  Could you make the change there as well?  ...and squash the two commits? :smile:

hymerman:
Don't worry about this now, I've just submitted a PR (#158) with the changes to `rcAlloc` too.

Having a quick tidy up of some low-hanging fruit :)

PR #66, opened 2015-01-02
Added Visual Studio 2013 idb files to gitignore.

grahamboree:


memononen:
Thanks for adding that!

PR #67, opened 2015-01-02
Replaced math function macros with inline functions.

grahamboree:
Also changed fabs reference to fabsf, as all callers are passing floats anyway and renamed dtMathFabs to reflect this.


memononen:
Excellent! That extra f sometimes confuses me.

PR #68, opened 2015-01-04
Removed redundant dtSqrt function and replaced calls to it with dtMathSqrtf

grahamboree:


Kromster80:
I was wondering, why were these wrapper functions introduced in the first place?

grahamboree:
Based on e5cca88cd116328d6eacdae6428614912c832850 it looks like they were introduced before DetourMath.h existed.

memononen:
Good riddance, thanks!

The wrapper was there to avoid including math.h in the header file for the inline vector functions.

PR #69, opened 2015-01-04
Changed straight path flags enum to a bitfield

grahamboree:
This shouldn't have any effect at all on the execution of the code, but it makes it a lot easier to read.  

Is the conversion at Detour/Source/DetourNavMeshQuery.cpp:1630 correct or should the original code have been `flags & DT_STRAIGHTPATH_END`?

### DIFF
--- a/Detour/Source/DetourNavMeshQuery.cpp
+++ b/Detour/Source/DetourNavMeshQuery.cpp
@@ -1605,8 +1605,8 @@ dtStatus dtNavMeshQuery::finalizeSlicedFindPathPartial(const dtPolyRef* existing
 }
 
 
-dtStatus dtNavMeshQuery::appendVertex(const float* pos, const unsigned char flags, const dtPolyRef ref,
-									  float* straightPath, unsigned char* straightPathFlags, dtPolyRef* straightPathRefs,
+dtStatus dtNavMeshQuery::appendVertex(const float* pos, dtStraightPathFlags flags, const dtPolyRef ref,
+									  float* straightPath, dtStraightPathFlags* straightPathFlags, dtPolyRef* straightPathRefs,
 									  int* straightPathCount, const int maxStraightPath) const
 {
 	if ((*straightPathCount) > 0 && dtVequal(&straightPath[((*straightPathCount)-1)*3], pos))
@@ -1627,7 +1627,7 @@ dtStatus dtNavMeshQuery::appendVertex(const float* pos, const unsigned char flag
 			straightPathRefs[(*straightPathCount)] = ref;
 		(*straightPathCount)++;
 		// If reached end of path or there is no space to append more vertices, return.
-		if (flags == DT_STRAIGHTPATH_END || (*straightPathCount) >= maxStraightPath)
+		if ((flags.straightpathEnd && !flags.straightpathStart && !flags.straightpathOffmeshConection) || (*straightPathCount) >= maxStraightPath)
 		{
 			return DT_SUCCESS | (((*straightPathCount) >= maxStraightPath) ? DT_BUFFER_TOO_SMALL : 0);
 		}
@@ -1636,8 +1636,8 @@ dtStatus dtNavMeshQuery::appendVertex(const float* pos, const unsigned char flag
 }
 
 dtStatus dtNavMeshQuery::appendPortals(const int startIdx, const int endIdx, const float* endPos, const dtPolyRef* path,
-									  float* straightPath, unsigned char* straightPathFlags, dtPolyRef* straightPathRefs,
-									  int* straightPathCount, const int maxStraightPath, const int options) const
+									   float* straightPath, dtStraightPathFlags* straightPathFlags, dtPolyRef* straightPathRefs,
+									   int* straightPathCount, const int maxStraightPath, const int options) const
 {
 	const float* startPos = &straightPath[(*straightPathCount-1)*3];
 	// Append or update last vertex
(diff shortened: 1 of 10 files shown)


Detour/Source/DetourNavMeshQuery.cpp
```diff
@@ -1827,11 +1829,11 @@ dtStatus dtNavMeshQuery::findStraightPath(const float* startPos, const float* en
 					dtVcopy(portalApex, portalLeft);
 					apexIndex = leftIndex;
 					
-					unsigned char flags = 0;
+					dtStraightPathFlags flags;
 					if (!leftPolyRef)
-						flags = DT_STRAIGHTPATH_END;
+						flags.straightpathEnd = true;
```
hymerman: This line and the similar ones below have also changed meaning; they used to also set all other flags to false. Not sure if it's an issue or not, I'm just pointing it out.

Detour/Source/DetourNavMeshQuery.cpp
```diff
@@ -1916,7 +1918,9 @@ dtStatus dtNavMeshQuery::findStraightPath(const float* startPos, const float* en
 		}
 	}
 
-	stat = appendVertex(closestEndPos, DT_STRAIGHTPATH_END, 0,
+	dtStraightPathFlags endflag;
+	endflag.straightpathEnd = true;
```
hymerman: This is clumsy. Perhaps a helper using the 'builder' pattern would tidy it up, allowing you to write e.g. `straightPathFlags().straightpathEnd(true)`. Or perhaps static constants for the values would be useful.

hymerman:
This kind of change makes me wish we had some kind of microbenchmark suite so we can be more certain we're not affecting performance :) The code does look cleaner in general though; nice.

Detour/Source/DetourNavMeshQuery.cpp
grahamboree: While `flags = DT_STRAIGHTPATH_END;` and `flags.straightpathEnd = true;` are indeed semantically different, in this context (where `flags` is initialized at line 1832) they're identical.

hymerman: Ah, my mistake - sorry!

grahamboree:
My reasoning on it being identical (or potentially better) performance:

* In VS2013 sizeof(unsigned char) == sizeof(dtStraightPathFlags), though this is implementation-defined.
* The compiler is free to change the ordering and alignment if it feels it would be faster one way than another.
* There's no indirection through getters/setters (though they'd be inlined anyway)

I agree, having a benchmark suite would be very useful for changes like these, though I doubt any performance hit would be measurable.

grahamboree:
Added a few functions to make creating single-flag instances a bit nicer.  e.g.

`dtStraightPathFlags.WithStraithPathEnd(true)`

memononen:
Hi! I appreciate the work you put into this, but the style is very different from the rest of detour code. The dtStatus suffered from the same problem, and i chose to use a couple of helper functions instead. Would that make the logic simpler?

--mikko

grahamboree:
@memononen are you referring to the "With..." functions, or the change in general?

memononen:
@grahamboree Yes :) Or more precisely, I meant that would a couple of helper functions make writing the logic simpler instead?
```C++
if (dtIsStraightPathEnd(flag) && ...)
   foo();
```
The helper function, while not as pretty, is consistent with the current codebase. There's a lot of code which uses flags (poly flags, query filter, etc), and I think that the proposed change in the PR is quite inconsistent wit the rest of the code. I do agree with the general idea that we should make using the flags easier to read and write logic.

grahamboree:
The problem I see is more with a lack of types in certain places.  For example:

```C++
class dtQueryFilter
{
...
	unsigned short m_includeFlags;		///< Flags for polygons that can be visited. (Used by default implementation.)
	unsigned short m_excludeFlags;		///< Flags for polygons that should not be visted. (Used by default implementation.)
...
```

The only way that you ever find out that this has anything at all to do with `dtPoly::flags` is to read through the implementation or infer it from the comment.  Sure, documentation can help, but just relying on documentation is a crutch.  `m_excludeFlags` is not a short, it just happens to be implemented as occupying two bytes.  Furthermore, if a user wanted to handle more than 8 polygon flags, they'd have to modify the Detour code to expand the size of that member and hope they didn't break anything in the process.  If it's just filtering polygons, it really should just be left to the user as a function to implement themselves, though that's beside the point.

The change is just a small step in this direction, (adding type information to non-typed values) but I think Recast would benefit from it.  If consistency is an issue, I could make progress on other flags as well in a separate branch and submit a single, large pull request when things look consistent.  I'm reluctant to take that approach, though, as merging large diffs when your code has no unit tests is pretty scary.

hymerman:
I agree that the current convention (integer types and bit manipulation) is not great. I don't think that bitfields are the best solution either though.

This kind of pattern is repeated throughout the codebase; perhaps a nice helper class would be a good idea? Something like one of the following (or a C++03 equivalent probably using macros):

https://gist.github.com/dscharrer/371337722bb56c45b008
http://blog.codef00.com/2014/12/06/portable-bitfields-using-c11/

... which could allow code that looks something like the following:

    enum class PolyFlag {
        Water,
        Road,
        flags_count_
    };
    typedef ::flags<PolyFlag> PolyFlags;
    
    class dtQueryFilter
    {
    ...
        PolyFlags m_includeFlags;
        PolyFlags m_excludeFlags;

There are lots of similar options and interfaces of course, these were just the first a quick google turned up.

PR #74, opened 2015-01-20
add ‘const ’ modifiers

niuys:
add ‘const ’ modifiers


hymerman:
Hey! Thanks for contributing - I love `const` correctness fixes!

Just a quick note on git etiquette for future contributions; it's normal to make commits self-contained and atomic (so a set of changes that depend on each other should go in one commit), and to put the description of the changes in the commit messages as well as the pull request (they'll probably be the same thing if it's a one-commit pull though). Pull requests are a github concept so pull request messages aren't available unless you're browsing the github website.

It's a shame that the github UI fills in a default commit message of "Update file.ext" since that's not very descriptive :)

niuys:
Thanks for your tip

PR #75, opened 2015-01-21
Replaced FileList class with std::vector<std::string>

grahamboree:
The goals of the demo project (as I see it at least) are to:
1.  Showcase Recast and Detour's capabilities and flexibility
2.  Provide a clear example of how to correctly interact with the Recast and Detour API's
3.  Provide a flexible base on which people can build proprietary tools.

The demo project doesn't have the same dependency restrictions as Detour and Recast, and benefits more from clarity and flexibility.  I feel like using STL is an improvement in both these areas over hand-rolled collections and c-style strings.  To be clear, I feel this only applies to the demo project, and not Recast and Detour as a whole.

I'd love to hear people's input on this, especially if they disagree.

### DIFF
--- a/RecastDemo/Source/SlideShow.cpp
+++ b/RecastDemo/Source/SlideShow.cpp
@@ -17,6 +17,7 @@
 //
 
 #include "SlideShow.h"
+
 #include <string.h>
 #include <stdio.h>
 #include "SDL_opengl.h"
@@ -98,7 +99,7 @@ void SlideShow::prevSlide()
 
 void SlideShow::setSlide(int n)
 {
-	const int maxIdx = m_files.size ? m_files.size-1 : 0;
+	const int maxIdx = m_files.size() ? m_files.size() - 1 : 0;
 	m_nextSlide = n;
 	if (m_nextSlide < 0) m_nextSlide = 0;
 	if (m_nextSlide > maxIdx) m_nextSlide = maxIdx; 
@@ -120,11 +121,11 @@ void SlideShow::updateAndDraw(float dt, const float w, const float h)
 	if (m_curSlide != m_nextSlide && m_slideAlpha < 0.01f)
 	{
 		m_curSlide = m_nextSlide;
-		if (m_curSlide >= 0 && m_curSlide < m_files.size)
+		if (m_curSlide >= 0 && static_cast<std::size_t>(m_curSlide) < m_files.size())
 		{
 			char path[256];
 			strcpy(path, m_path);
-			strcat(path, m_files.files[m_curSlide]);
+			strcat(path, m_files[m_curSlide].c_str());
 			loadImage(path);
 		}
 	}
--- a/RecastDemo/Source/main.cpp
+++ b/RecastDemo/Source/main.cpp
@@ -28,6 +28,9 @@
 #	include <GL/glu.h>
 #endif
 
+#include <vector>
+#include <string>
+
 #include "imgui.h"
 #include "imguiRenderGL.h"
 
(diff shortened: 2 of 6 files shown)


RecastDemo/Source/SlideShow.cpp
```diff
@@ -120,11 +120,11 @@ void SlideShow::updateAndDraw(float dt, const float w, const float h)
 	if (m_curSlide != m_nextSlide && m_slideAlpha < 0.01f)
 	{
 		m_curSlide = m_nextSlide;
-		if (m_curSlide >= 0 && m_curSlide < m_files.size)
+		if (m_curSlide >= 0 && (uint)m_curSlide < m_files.size())
```
hymerman: uint should be std::size_t

RecastDemo/Source/main.cpp
```diff
@@ -686,15 +686,16 @@ int main(int /*argc*/, char** /*argv*/)
 				mouseOverMenu = true;
 			
 			int levelToLoad = -1;
-			for (int i = 0; i < files.size; ++i)
-			{
-				if (imguiItem(files.files[i]))
+			
+			for (unsigned int i = 0; i < files.size(); ++i) {
```
hymerman: If I'm being picky, this unsigned int (and the one on line 768) should also be std::size_t to avoid warnings on 64 bit platforms, but that'll trigger all over the place already so don't worry!

hymerman:
I'm all for this kind of thing.

RecastDemo/Source/main.cpp
grahamboree: If you really want to be pedantic, the value returned by files.size() is a `std::vector<std::string>::size_type` value, but who wants to write that? :P

grahamboree:
To be ultra-pedantic, files.size() returns a std::vector<std::string>::size_type not a std::size_t.  It's almost certainly true that vector's size_type is indeed a size_t, but given the new diffs, the point is moot.  I changed it to use iterators which is more portable than indexing anyway.

hymerman:
Touché! That's where I draw the line :)

grahamboree:
@hymerman @memononen @Janiels What do you guys think about using the STL in RecastDemo?  I feel like it would make the code clearer and more robust.

I can understand and agree with avoiding STL in Recast and Detour.  However, I don't think people would be using RecastDemo code in their game runtime, or care if a tool in their production pipeline uses the STL or not.

hymerman:
I'm still all for it :) In fact I'd also use it in Recast and Detour if it was purely up to me but I know a lot of people feel passionately that that's a bad idea so I'll stay (relatively) quiet!

One thing I'd request before accepting this PR though: no `using` in headers.

Janiels:
I think that's a good idea too. The point of RecastDemo is to showcase and exemplify the library and I don't think that has to be obfuscated by too many manual std::vector implementations. :smile:
Definitely think that R&D itself should stay low-level, however, given some of the constraints of the users.

grahamboree:
@Janiels I agree.  While it'd be nice to have the STL in R&D, I think it's not enough of a reason to force users to link against it.

Also, at some point we really should have a C API to allow exposing FFI or DLL bindings.  Using the STL in R&D would make this more difficult, though not impossible.

Anyway, I removed usings from the headers, (good catch @hymerman :stuck_out_tongue:) so this should be good to go.

Janiels:
LGTM, but I'll let @hymerman merge.

hymerman:
Done! Thanks :)

memononen:
At the time of writing Recast there was this movement of no-alloc and "std sucks", which I embraced :) Not so strict anymore, though. I would recommend against std in the core lib unless you have a good way of implement custom allocators. Most studios I know use their own allocators.

PR #76, opened 2015-01-21
Removed a smattering of compiler warnings

grahamboree:
Unused variables, signed/unsigned mismatching, and a global variable.  Also changed test case include/exclude flags to unsigned shorts to match their usage,

### DIFF
--- a/DetourCrowd/Source/DetourObstacleAvoidance.cpp
+++ b/DetourCrowd/Source/DetourObstacleAvoidance.cpp
@@ -331,7 +331,7 @@ float dtObstacleAvoidanceQuery::processSample(const float* vcand, const float cs
 	// find the threshold hit time to bail out based on the early out penalty
 	// (see how the penalty is calculated below to understnad)
 	float minPen = minPenalty - vpen - vcpen;
-	float tThresold = ((double)m_params.weightToi/(double)minPen - 0.1) * (double)m_params.horizTime;
+	float tThresold = (m_params.weightToi / minPen - 0.1f) * m_params.horizTime;
 	if (tThresold - m_params.horizTime > -FLT_EPSILON)
 		return minPenalty; // already too much
 
@@ -528,7 +528,6 @@ int dtObstacleAvoidanceQuery::sampleVelocityAdaptive(const float* pos, const flo
 	
 	const int nd = dtClamp(ndivs, 1, DT_MAX_PATTERN_DIVS);
 	const int nr = dtClamp(nrings, 1, DT_MAX_PATTERN_RINGS);
-	const int nd2 = nd / 2;
 	const float da = (1.0f/nd) * DT_PI*2;
 	const float ca = cosf(da);
 	const float sa = sinf(da);
--- a/RecastDemo/Source/TestCase.cpp
+++ b/RecastDemo/Source/TestCase.cpp
@@ -146,7 +146,7 @@ bool TestCase::load(const char* filePath)
 			test->expand = false;
 			test->next = m_tests;
 			m_tests = test;
-			sscanf(row+2, "%f %f %f %f %f %f %x %x",
+			sscanf(row+2, "%f %f %f %f %f %f %hx %hx",
 				   &test->spos[0], &test->spos[1], &test->spos[2],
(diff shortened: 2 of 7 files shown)


RecastDemo/Source/TestCase.cpp
```diff
@@ -199,7 +199,7 @@ void TestCase::doTests(dtNavMesh* navmesh, dtNavMeshQuery* navquery)
 		iter->nstraight = 0;
 		
 		dtQueryFilter filter;
-		filter.setIncludeFlags((unsigned short)iter->includeFlags);
+		filter.setIncludeFlags(iter->includeFlags);
```
hymerman: Typo? Seems weird that this line has changed but the one below hasn't.

grahamboree: Nice catch, thanks!

DetourCrowd/Source/DetourObstacleAvoidance.cpp
```diff
@@ -331,7 +331,7 @@ float dtObstacleAvoidanceQuery::processSample(const float* vcand, const float cs
 	// find the threshold hit time to bail out based on the early out penalty
 	// (see how the penalty is calculated below to understnad)
 	float minPen = minPenalty - vpen - vcpen;
-	float tThresold = ((double)m_params.weightToi/(double)minPen - 0.1) * (double)m_params.horizTime;
+	float tThresold = (m_params.weightToi / minPen - 0.1f) * m_params.horizTime;
```
grahamboree: I'm not sure why these were being cast to double before the division. Maybe it was done in an attempt to increase the precision of the division?  Unless there's something I'm missing, this is generally not the case for IEEE 754 floats.

hymerman: Oh my. This comment just made me re-read this (brilliant) article:

https://randomascii.wordpress.com/2012/03/21/intermediate-floating-point-precision/

So... maybe the casts were a good idea, maybe they weren't. Maybe it depends on how the code is compiled, and a hundred other things. I think it's probably best to keep the behaviour the same if we don't know what the impact will be - and just cast the original back to float explicitly.

grahamboree: Nice article!  I definitely learned a bit about FLOPs.

It looks like 99d1e791ab9898af18eb08b1e5a7983c5065143e introduced this.  Maybe @axelrodR can help clarify the intent here?

axelrodR: I am not sure there is any reason to the cast there, but I don't really remember. Rethinking the subject, all the parameter values composing that formula are not very sensitive to mantissa size and have similar exponent:
 - m_params.weightToi, m_params.horizTime: The numeruical values that make sense are in the range of 0.5-5 (2.5 in the demo). 
 - minPen is a sum of the current minPenalty and several normalized difference in velocity (normalized is key here), so it either float max or in the range of 0-20, with large jumps between values due to angular change in sampled velocity.

 

    On Monday, January 11, 2016 8:47 PM, Graham Pentheny <notifications@github.com> wrote:
 

 In DetourCrowd/Source/DetourObstacleAvoidance.cpp:> @@ -331,7 +331,7 @@ float dtObstacleAvoidanceQuery::processSample(const float* vcand, const float cs
>  	// find the threshold hit time to bail out based on the early out penalty
>  	// (see how the penalty is calculated below to understnad)
>  	float minPen = minPenalty - vpen - vcpen;
> -	float tThresold = ((double)m_params.weightToi/(double)minPen - 0.1) * (double)m_params.horizTime;
> +	float tThresold = (m_params.weightToi / minPen - 0.1f) * m_params.horizTime;
Nice article! I definitely learned a bit about FLOPs.It looks like 99d1e79 introduced this. Maybe @axelrodR can help clarify the intent here?—

hymerman: Cool - sounds like it should be fine to make the change then.

hymerman:
I think the rest of this PR is good to go; I'll merge now.

PR #77, opened 2015-01-26
Possible new function for recast

JimmyJames707:
Hi Mikko,
We discussed adding this function some time ago. This post has the details:

https://groups.google.com/forum/#!topic/recastnavigation/wj_4UnOhXFY

### DIFF
--- a/DetourCrowd/Include/DetourCrowd.h
+++ b/DetourCrowd/Include/DetourCrowd.h
@@ -343,6 +343,17 @@ public:
 
 	/// Gets the query object used by the crowd.
 	const dtNavMeshQuery* getNavMeshQuery() const { return m_navquery; }
+	
+	/// Finds all agents within range of centerPos, excluding the specified agent.
+	///  @param[in]		centerPos		The center of the search circle. [(x, y, z)]
+	///  @param[in]		height			The height of the search circle.
+	///  @param[in]		radius			The radius of the search circle.
+	///  @param[in]		skipIdx			The optional index of the agent to be excluded from the query. [Limits: 0 <= value < #getAgentCount()] -1 ignore
+	///  @param[out]	crowdNeighbour	An array of crowd neighbour pointers. [(#dtCrowdNeighbour *) * maxNeighbours]
+	///  @param[in]		maxNeighbours	The maximum size of the crowd neighbor array.
+	///  @param[out]	numNeighbours	The number of crowd neighbors found
+	/// @return True if the request was successfully submitted.
+	bool findAgentsAroundCircle(const float *centerPos, const float height, const float radius, int skipIdx, dtCrowdNeighbour *crowdNeighbour, const int maxNeighbours, int *numNeighbours);
 };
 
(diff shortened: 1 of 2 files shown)


leif81:
Would it suffice if the existing static function `getNeighbours` were made public instead?

JimmyJames707:
Wouldn't bother me, but getAgentIndex would also have to be made public.

grahamboree:
I'm much more in favor of making the existing function public than duplicating functionality.  @memononen @hymerman @Janiels Do you guys see any reason that exposing these functions would be problematic?  Maybe renaming "getNeighbors" to something less usage-specific like "getAgentsInRange" would also be a good change?

Janiels:
Hmm, I'm not completely sure. `getNeighbours` has a pretty complicated signature, and if the common functionality is to "find agents in a circle", I think we should rather expose that as a simpler API. `getNeighbours` and `getAgentIndex` look like they are pretty "low level". I think we should attempt to keep the surface area of the public API as small as possible - so unless there are more usage cases for those methods rather than just `findAgentsAroundCircle`, I don't think we should make those public.

I'm not completely sold on the implementation in this PR though. I feel `skipIdx` does not really belong in this API; seems like pollution to me, compared to just doing that filtering yourself after obtaining the results. Especially if the function is named `findAgentsAroundCircle`. Also, if we take the array of `dtCrowdNeighbour`, I think we should fill the distances too. Otherwise we should just take an `int` array for the integers.

hymerman:
Seems to me there are two things that users could want to with this; find neighbours of an agent, and find agents close to a position. Perhaps it would be best to have two functions; a friendlier `getNeighbours` and `findAgentsAroundCircle`.

`findNeighbours` would take an agent (which is not returned as a neighbour), and doesn't take a position (instead using the agent's position), and doesn't take `skipIdx` as that's implicitly the agent passed in.

`findAgentsAroundCircle` would be for cases where the caller does want to just find "all agents within range of this position", and would take a position but no `skipIdx`. Taking `skipIdx` implies that it's being used to find neighbours of an agent (and the caller doesn't consider an agent a neighbour of itself).

So yeah, I think remove `skipIdx` from this new function, and we should also add a friendlier `getNeighbours`. And @Janiels I agree, perhaps we should consider not exposing `dtCrowdNeighbour; returning an array of indices and an (optional?) array of distances instead would be in line with the rest if the Detour API.

grahamboree:
@hymerman That API description sounds a lot better than just exposing this function.  They're two separate but related tasks, and un-conflating them is probably best in the long run for both clarity and ease of use.

memononen:
I also agree that two separate functions would be best. One thing that hold me back implementing the neighbour access is that the data returned is not safe (IIRC they are indices to list of active agents).

Before adding the functionality, I suggest to change the public API to use a salted handles (similar to polygons) for the agents. That makes it safe to return handles to agents on query methods. This also means that `addAgent()`needs API breaking change, might not be a big deal, though.

PR #79, opened 2015-02-01
change DebugUtils/Source/DetourDebugDraw.cpp no executable

mynz:
DebugUtils/Source/DetourDebugDraw.cpp should be no executable.


grahamboree:
LGTM

PR #84, opened 2015-02-21
Remove vertex overrun fix

Kromster80:


hymerman:
I don't think this looks right - it doesn't look like the code reads outside the buffer anyway, and this change will make it not copy enough vertices. Are you sure this is correct? Do you have a case where the buffer is over-read?

Kromster80:
Sorry, I must have formulated my commit message wrong.

The commit fixed the unneeded vertex copy.

hymerman:
I've not studied the surrounding code that closely but assuming that loop is the only bit responsible for effectively removing the vertex with inded `rem`, by shuffling everything above it down one, then surely the last vertex isn't going to get copied now? Say you had only two vertices (I know you wouldn't but it's the easiest case to illustrate) and you wanted to remove vertex 0; the loop won't execute at all, nothing will get copied, so after the 'remove' you'll actually be left with just old vertex 0, not old vertex 1. Am I wrong?

Kromster80:
Oh, sorry, silly me! I have blindly pulled this change twice, each time just doing the -1. 
Of course this one is mistake! Thanks for noticing and pointing!
The correct PR was accepted 2 days ago.

hymerman:
No problem! I was starting to think I'd forgotten C++ for a minute ;)

PR #86, opened 2015-03-14
Add support to use GLFW (linux only for now);

fungos:
This implements a very thin and (ugly?)hacky layer that makes glfw behave as SDL, so to avoid a major rewrite of RecastDemo main.

I've added a premake option "--use-glfw" with a configuration for linux only (for now), adding support to windows and macosx should be trivial and need only changes to the premake file. I'm not doing this right now due to not being able to test it on these platforms.

Also note that the premake file changes may not be optimal, I still do not have a good premake knowledge.

Why GLFW?

Firstly because I've [implemented](https://github.com/fungos/fips-recast) support to [fips build system](https://github.com/floooh/fips) and at this moment we haven't a fipsified SDL build, only GLFW. 
But also, GLFW is an easier, good and lightweight alternative to SDL.

Why FIPS?

Fips is an amazing build system based on CMake where it manage all your (fips) dependencies automaticaly, having more and more fipsified projects means lower friction to programmers when adding and maintaining thirdparty dependencies. For example, commonly used "contrib" directories with instructions, git submodules, or just duplicated dependencies aren't needed anymore. Also, users using projects with this kind of dependencies will have not to worry anymore about Doing The Right Thing(tm).

### DIFF
--- a/RecastDemo/Include/wnd.h
+++ b/RecastDemo/Include/wnd.h
@@ -1,0 +1,108 @@
+#ifndef WND_H
+#define WND_H
+
+#if !defined(USE_GLFW)
+
+#include "SDL.h"
+#include "SDL_opengl.h"
+
+#define WND_QUIT                    SDL_QUIT
+
+#define WND_KEY_1                   SDLK_1
+#define WND_KEY_9                   SDLK_9
+#define WND_KEY_0                   SDLK_0
+#define WND_KEY_W                   SDLK_w
+#define WND_KEY_S                   SDLK_s
+#define WND_KEY_A                   SDLK_a
+#define WND_KEY_D                   SDLK_d
+#define WND_KEY_T                   SDLK_t
+#define WND_KEY_TAB                 SDLK_TAB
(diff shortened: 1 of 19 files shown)


Kromster80:
I'm not sure I understand the logic behind it. Why there is a need replace SDL solution with GLFW in the first place?

fungos:
This does not replace. GLFW is an alternative and optional.
The reasoning is described in the PR, though it may not be enough for some people.

TL;DR: 1) for use with a better build system than premake4 and; 2) GLFW for this kind of demo is better than SDL.

memononen:
I like GLFW, I've used in most of my other projects. The right solution would be to porte the demo to use GLFW. GLFW also pretty much requires GLEW in windows, you can take a look at nanovg on how to set that up properly [1]. Supporting the library takes considerable amount of my free time and I dont want to maintain a hack.

[1] https://github.com/memononen/nanovg

fungos:
Nice to hear you're open to GLFW. So I will do a clean port as I too do think is the right solution.

PR #88, opened 2015-03-18
RecastDemo conversion from SDL to GLFW3

fungos:
This also removes GLU and adds a dependency on GLEW.

The reasoning for this is the hability of using a new build system, [fips](http://github.com/floooh/fips), that manages dependencies automaticaly on any platform. But also GLFW is a more lightweight library and generally easier dependency to manage for Windows and MacOS users.


grahamboree:
So, this PR is pretty old.  

The proposed rationale is that it's easier to dependency manage on different platforms and compatibility with a different build system.  IMO setting up SDL is very straightforward on Linux OSX and Windows.  There has been some discussion on the mailing list about switching build systems, but the conclusion was to stick to premake5.

SDL is backed by big players like Valve and Double Fine, and (from what I've heard) has better support for different platforms and game controllers than GLFW.  I think it's good that there's a fork of recast with a GLFW backend, but I don't know if there are enough compelling reasons presented to merge it into the official repo.

hymerman:
Agreed. We should close this. Sorry @fungos! Thanks for the PR anyway!

PR #89, opened 2015-03-23
Continuous integration using AppVeyor on Windows and Travis on Linux

hymerman:
A while ago we were talking about getting automated builds set up, so here's one for Windows!

It's using AppVeyor (http://www.appveyor.com/) to build on Windows, with VS2013 and VS2015 CTP 6, in Release and Debug configs. I've included premake5 in the repository (I personally like having the build tool versioned, just shout if that's not your style). The script downloads and extracts SDL, and uses premake to build the VS projects.

I had to use premake 5 as I haven't been able to get AppVeyor to build with anything before VS2013, which premake 4 doesn't support.

I've also added a little status badge to the readme, above the Bitdeli badge.

You can see it in action here:
https://ci.appveyor.com/project/hymerman/recastnavigation

After merging this, you'll have to do a few things that I can't do:
* Sign up for AppVeyor (just log in with GitHub)
* Add recastnavigation as a project (just click 'new project', 'github', then select from your projects)
* Change the badge in the readme to point to your appveyor project instead of mine - in appveyor go to settings -> badges and find the ID embedded in any of them, just replace my id with yours

Future work:
* Extend this to build x64 and x86 configs after they've been added to the premake file
* Figure out how to do pull request hooks to tell you at a glance if pull requests pass the build
* Simple automated tests with UnitTest++ (or some other if you prefer)
* All of this on Linux with Travis CI


hymerman:
Oh, the badge change will also need you to change 'hymerman' to 'memononen' in the link, clearly!

hymerman:
Just added binaries for Linux and Mac premake5 and updated documentation.

hymerman:
On the subject of pull request hooks: looks like that's something entirely managed by the integration, so Travis already does it (and will start adding notes to PRs), and AppVeyor might do it too, not sure.

hymerman:
I've added a Travis CI config which seems to be working (hard to tell since Travis is having some weird issues right now). See it in action here: https://travis-ci.org/hymerman/recastnavigation

You'll need to:
 * Sign up for Travis (I think you can use your GitHub account again)
 * Tell it to build recastnavigation
 * In the project settings, turn the three options on (build pushes, build pull requests, only build if .travis.yml is present)

... and again you'll want to change the badge URL in README.md to point to yours, not mine, though this time it's just names and no ids so it's easier.

Note that building with clang is disabled since there seems to be some issue with fast fp math (see the errors here: https://travis-ci.org/hymerman/recastnavigation/jobs/56234581 and this bug: https://llvm.org/bugs/show_bug.cgi?id=13745).

This is a separate issue, but is fast fp math really required? I've found it to be more trouble that it's worth on other projects, and was using Recast/Detour without it for Alien: Isolation.

hymerman:
I'll wait until this is merged before starting on unit testing and x64 builds to avoid having to stick it all in one massive pull request :)

grahamboree:
While extremely convenient to have, I'm not sure we should have the premake binaries included in the project.  It bloats the repo size, (git doesn't deal with binary files well) but more importantly, it's a source for confusion.  I currently have premake installed in "/usr/local/bin" and having it in the project might confuse users with a similar setup into thinking it's some recast-customized version that they should use.  Instructions on installing premake have been in the readme for a long time, and people seem okay with having to install it themselves.
Regarding CI, it's possible that Travis and Apveyor already have premake installed, though we can have the scripts download is as part of a build step if that's not the case.

hymerman:
Yeah, I'm in two minds about this kind of thing. I do it in my own projects with small tools to make CI iteration time faster for small builds.

In fairness, it's less than 400KB (I've seen bigger source files!). Also at the time I wrote this, I upgraded to premake5 to allow building of VS2013 and VS2015 projects, and the premake project had only just added their first built release to github, so I wasn't sure if downloading it in the build script would be a good idea since it was still in flux.

But anyway, that's all ancient history! Travis doesn't have premake installed (https://docs.travis-ci.com/user/ci-environment/). Perhaps now it would be a good idea for the build script to download the latest release from https://github.com/premake/premake-core/releases? I'll give this a try and update the PR later.

Or... CMake seems to be more popular these days - maybe it's worth considering using it instead of premake?

leif81:
I too would suggest avoid committing build tools into the project.

@hymerman I like your idea of downloading build dependencies (i.e. premake binary) dynamically as part of the build.

grahamboree:
Mmm I think I just don't like it out of principle. :stuck_out_tongue_winking_eye:  I'm really not too familiar with CMake - I've used it, but never really tried to set it up for my own project before.  I think just doing a wget for the premake binary is fine until there's a more compelling reason to spend more effort on it.
For the record, I'm not against switching to cmake if it makes things easier.  Cmake is also a bit more pervasive, so it's possible that people are more familiar and comfortable using it?  Other than that I don't really know of any enormous benefits one way or the other.

hymerman:
Closing; superseded by #131

PR #90, opened 2015-03-24
Make max number of convex volumes configurable

leif81:
This removes the arbitrary max upper limit of 256 in case you need a very large number.

The background for this change is that @JeanPierreC and I ran into an issue with a nav mesh we were building that required several thousand convex volumes to represent our road network. We had been using `InputGeom.h/cpp` directly from the _RecastDemo_ in our project and this is the fix we required. We had originally just tried increasing `MAX_VOLUMES` to what we needed (32768), but that blew the 1MB stack size pretty easily. So instead we went with this solution to dynamically allocate space for the volumes on the heap. We can now can guarantee there'll be exactly enough convex volumes rather the hard coded limit to bump into.

This change shouldn't be noticable to the _RecastDemo_, the default number of volumes is still 256, but we thought if someone else is like us and uses the _RecastDemo_ source code as the basis for their project this may be useful to them.

### DIFF
--- a/RecastDemo/Include/InputGeom.h
+++ b/RecastDemo/Include/InputGeom.h
@@ -51,13 +51,13 @@ class InputGeom
 
 	/// @name Convex Volumes.
 	///@{
-	static const int MAX_VOLUMES = 256;
-	ConvexVolume m_volumes[MAX_VOLUMES];
+	ConvexVolume * m_volumes;
+	const int m_maxVolumes;
 	int m_volumeCount;
 	///@}
 	
 public:
-	InputGeom();
+	InputGeom(int maxvolumes = 256);
 	~InputGeom();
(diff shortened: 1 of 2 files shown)


Janiels:
I would prefer if it was changed to `std::vector` so it can grow if required; it also makes the code a bit cleaner. Can you make that change instead?

Sorry for the very late reply.

leif81:
Sure I can do that.

leif81:
FYI, this is still on my radar. I've just been working on other projects and haven't had recastnavigation dev environment installed in a while.

hymerman:
:+1: Thanks :)

jakobbotsch:
@leif81 I will close this for now. Please reopen if you do end up making that update.

PR #91, opened 2015-04-26
RecastDemo Allow the camera to be rotated on the X axis using Q and E.

Warpten:
This patch just allows to alter camera yaw using Q and E in complement of the mouse. I am not sure which keys to use to alter pitch. I tried X and space, but space would conflict with previously defined actions (that I personally do not need). Let me know what you think of it.


grahamboree:
I don't think this functionality is really necessary.  If anything I think we'd want to copy the controls of something like [this](http://docs.unity3d.com/Manual/SceneViewNavigation.html).  Also, this change is not compatible with the recent switch to SDL2.

PR #93, opened 2015-04-28
Remove dead assignments

rafzi:
This PR removes dead assignments. It also includes one style fix and one typo fix.

The pointer increments are left in a comment in case they are needed anytime in the future.

Those errors were found with the llvm scan-build utility. It also found a couple divisions by zero that result from the assumption that dtPoly::vertCount is always greater than zero. I assume this is a performance critical assumption. The following was also found, but I do not understand the code well enough to judge if this is a valid bug scenario:

In RecastMeshDetail.cpp:
In triangulateHull the statements at "Add first triangle" seem to pass uninitialized values in the following condition:
- called from buildPolyDetail
- sampleDist <= 0
=> hull is uninitialized

### DIFF
--- a/Detour/Source/DetourNavMesh.cpp
+++ b/Detour/Source/DetourNavMesh.cpp
@@ -304,7 +304,7 @@ int dtNavMesh::findConnectingPolys(const float* va, const float* vb,
 	if (!tile) return 0;
 	
 	float amin[2], amax[2];
-	calcSlabEndPoints(va,vb, amin,amax, side);
+	calcSlabEndPoints(va, vb, amin, amax, side);
 	const float apos = getSlabCoord(va, side);
 
 	// Remove links pointing to 'side' and compact the links array. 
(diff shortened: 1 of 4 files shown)


Detour/Source/DetourNavMesh.cpp
```diff
@@ -920,7 +920,7 @@ dtStatus dtNavMesh::addTile(unsigned char* data, int dataSize, int flags,
 	tile->detailVerts = (float*)d; d += detailVertsSize;
 	tile->detailTris = (unsigned char*)d; d += detailTrisSize;
 	tile->bvTree = (dtBVNode*)d; d += bvtreeSize;
-	tile->offMeshCons = (dtOffMeshConnection*)d; d += offMeshLinksSize;
+	tile->offMeshCons = (dtOffMeshConnection*)d; // d += offMeshLinksSize;
```
grahamboree: IMO I think this dead-store makes things more clear to the reader that we're just reading data from a byte buffer and advancing a read pointer.  Additionally, it makes the code a bit more robust to changes, because you don't have to figure out if you need to un-comment some code that you didn't write.

hymerman:
I'm not going to merge this as-is; I've just had a look and I think some of those `stat` assignments should be checked (rather than the assignment removed) and I'll write a nice `getThenAdvanceBufferPointer` function to make the intent of the 'moving through buffer sections' code clearer (and fix the warnings while I'm at it!). I'll credit this PR with the other changes of course!

Out of interest, is there a nice simple way to run scan-build with our premake build? Was it hard to set up? I'd love to run this myself!

rafzi:
Alright, I hope this helped. Keeping it open as open issue.

I never used premake before, but scan-build should work with absolutly everything. Just prepend it to all build commands so it can register itself as compiler. So it should be something like:

```
scan-build premake xxx
scan-build make
```

hymerman:
Oh cool! Nice and easy :) I'll give it a go soon - thanks.

Yes, let's leave this open just to track it, I'll close once I've submitted my changes.

hymerman:
Just FYI, I've fixed all these issues but in another way, in PR #192. Thanks for pointing them out!

Going forward, I'm going to use scan-build and Coverity Scan regularly to keep warnings low.

PR #94, opened 2015-05-01
reset agent desired velocity in resetMoveTarget

leif81:
Otherwise, if velocity was non zero at time resetMoveTarget is called the agent would continue to move indefinitely at that velocity (a straight line) which was not the intended behaviour.

As discussed in https://groups.google.com/forum/#!topic/recastnavigation/QoLNGz5o6e8


memononen:
Looks good, thanks for the fix!

leif81:
:tada: Terrific, thank you for the merging @memononen !

PR #97, opened 2015-05-11
Fix for wrongly simplified neighbour contours

carlesros:
Two fixes for walkContour/simplifyContour:
- Fix for floating error innacuracy when calculating the maximum
deviation during simplifyContour: ensure points are passed to
distancePtSeg on the same order for both neighbour contours.
- Fix for regions touching only in one vertex: during walkContour mark
vertices that touch another region in only one vertex, and then let
simplifyContour insert that vertex in the initial polygon.

See the post "Two bugs in contour generation (and fixes)" in Recast
forum for a detailed description
https://groups.google.com/forum/#!topic/recastnavigation/OHK5JWLvWs0

### DIFF
--- a/DetourTileCache/Source/DetourTileCacheBuilder.cpp
+++ b/DetourTileCache/Source/DetourTileCacheBuilder.cpp
@@ -488,6 +488,66 @@ static bool walkContour(dtTileCacheLayer& layer, int x, int y, dtTempContour& co
 		
 		if (rn != layer.regs[x+y*w])
 		{
+			//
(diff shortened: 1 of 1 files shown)


Sandern:
It seems the changes assume dtTempContour::verts are unsigned shorts, rather than unsigned chars. Mistake, or am I missing something?

TheMartin:
Is there a chance this could be ported into the version of walkContour and simplifyContour that's being called from rcBuildContours?

PR #103, opened 2015-06-28
Fix 2 "unused parameter" build warnings

jackpoz:
Comment unused parameters to avoid build warnings when Recast and Detour and built with /W4 on windows or -Wunused-parameter on gcc/clang


jackpoz:
Thank you for merging the PR

Janiels:
Thanks for the fixes! :+1:

PR #108, opened 2015-08-18
Update DetourNavMeshQuery.cpp

huanzai:
fix the Segment fault when "nearestRef" is NULL

### DIFF
--- a/Detour/Source/DetourNavMeshQuery.cpp
+++ b/Detour/Source/DetourNavMeshQuery.cpp
@@ -711,8 +711,6 @@ dtStatus dtNavMeshQuery::findNearestPoly(const float* center, const float* exten
 										 dtPolyRef* nearestRef, float* nearestPt) const
 {
 	dtAssert(m_nav);
-
-	*nearestRef = 0;
 	
 	// Get nearby polygons from proximity grid.
 	const int MAX_SEARCH = 128;


Detour/Source/DetourNavMeshQuery.cpp
```diff
@@ -711,8 +711,6 @@ dtStatus dtNavMeshQuery::findNearestPoly(const float* center, const float* exten
 										 dtPolyRef* nearestRef, float* nearestPt) const
 {
 	dtAssert(m_nav);
-
-	*nearestRef = 0;
```
memononen: `nearestRef` should be 0 in case of failure below. Better solution would be:
```
 	if (nearestRef)
 		*nearestRef = 0;
```

Janiels:
I don't see how the method is useful without passing the `nearestRef` param; it will always return `DT_SUCCESS`, and it is not possible to detect if `nearestPt` has been updated in this situation.

What is the use case for passing `nullptr` here? I'm thinking it may be better to return `DT_FAILURE | DT_INVALID_PARAM` when it is null, instead of skipping the assignments.

hymerman:
The caller might just want `nearestPt` and not `nearestRef`. Mikko is right, the change should check `nearestRef` for `nullptr` before assigning... *if* we want to zero it out.

I'd question whether it's useful to zero out the ref at all; in general, callers shouldn't assume any 'out' parameters are valid if a function returns `DT_FAILURE`. They should be checking the return status before using results.

I can't remember too clearly, it's been a while since I was in the guts of Detour; what's the most common approach in the codebase right now, do we set valid 'zeroed' outputs on failure, or do we leave outputs undefined? If it's the former, we should use Mikko's suggestion, if it's the latter, we should accept the PR as it is.

jswigart:
I agree with not adopting initialization standards into out parameter. Either the function succeeds and the output parameters are valid, or the function fails and the output parameters should be untouched imo.

Janiels:
I think it's a little iffy since it's impossible to detect if `nearestPt` was assigned without checking `nearestRef` (well not impossible, but it requires some distance checks - surely just prettier to check `nearestRef` then).

Given this, would it really be useful to pass only `nearestPt`? The behavior is then "adjust the point to the closest polygon, or do nothing if there is no polygon close enough". Maybe that could be useful when snapping agents to mesh, I suppose.

Also, the function cannot return anything other than `DT_SUCCESS` at the moment - `queryPolygons` cannot fail. That's not clear from reading `findNearestPoly` alone. So in all cases `nearestRef` is assigned, and `DT_SUCCESS` is returned.

hymerman:
That's a fair point, yes - no nearby polys is a 'softer' fail (even if the proper fail can't happen currently) and is only detectable through `nearestRef`. In that case `nearestRef` isn't optional and should be asserted on, and comments should mention this. (even better would be to use a reference but that's something I'd like to do in one big breaking swoop for a big version bump).

Janiels:
Do we really want to assert on parameters in publicly exposed APIs, instead of checking them and returning `DT_FAILURE | DT_INVALID_PARAM`?
I get that for internal functions, but for APIs I think we should try to stay away from preconditions like that.

hymerman:
Ideally it'd be a reference, and the closest to a reference is a pointer + assert... but I'm happy either way to be honest. More error checking is good, whichever form it takes!

Janiels:
Ah yes, a reference would definitely be ideal. That clearly communicates that the parameter is required.
I made a PR for until we make that change in #136. Closing this since it is not relevant anymore.

PR #109, opened 2015-08-29
Updated dtFindPathOptions description and options

Kromster80:
"Options for dtNavMeshQuery::FindPath" is not correct, the option is used in SlicedFindPath only
DT_FINDPATH_LOW_QUALITY_FAR is not used anywhere

I would also add, that even though raycast considers costs, does so not very well :-(

### DIFF
--- a/Detour/Include/DetourNavMesh.h
+++ b/Detour/Include/DetourNavMesh.h
@@ -118,11 +118,10 @@ enum dtStraightPathOptions
 };
 
 
-/// Options for dtNavMeshQuery::findPath
+/// Options for dtNavMeshQuery::initSlicedFindPath and updateSlicedFindPath
 enum dtFindPathOptions
 {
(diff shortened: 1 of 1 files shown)


grahamboree:
Looks good to me.  Considering this was also marked with `[provisional]` I think it's safe to remove.

jackpoz:
according to https://github.com/recastnavigation/recastnavigation/pull/17#discussion_r11897066 the idea was to add the config first and then implement it properly, that was the idea behind marking it as "provisional".

instead of removing the option it should have been implemented

grahamboree:
That makes sense.  I didn't know about that PR comment.  I've added a task (#142) to actually implement this feature, but for now I'd rather not have the flag if it doesn't do anything.

Janiels:
I agree that we should not have the flag if it doesn't do anything. Sounds like a case of YAGNI to me.

PR #110, opened 2015-09-01
Comments

Kromster80:
Code comments as per https://groups.google.com/forum/#!searchin/recastnavigation/nvel/recastnavigation/xpHL5dJi538/olCPU3RxWNYJ

### DIFF
--- a/DetourCrowd/Include/DetourCrowd.h
+++ b/DetourCrowd/Include/DetourCrowd.h
@@ -143,9 +143,9 @@ struct dtCrowdAgent
 	float desiredSpeed;
 
 	float npos[3];		///< The current agent position. [(x, y, z)]
-	float disp[3];
-	float dvel[3];		///< The desired velocity of the agent. [(x, y, z)]
-	float nvel[3];
+	float disp[3];          ///< "disp" is just something used during collision detection
+	float dvel[3];		///< The desired velocity of the agent. [(x, y, z)] (based on the current path, calculated each frame "from scratch")
+	float nvel[3];          ///< Desired velocity adjusted by obstacle avoidance, calculated each frame "from scratch"
 	float vel[3];		///< The actual velocity of the agent. [(x, y, z)]
 
 	/// The agent's configuration parameters.


DetourCrowd/Include/DetourCrowd.h
```diff
@@ -143,9 +143,9 @@ struct dtCrowdAgent
 	float desiredSpeed;
 
 	float npos[3];		///< The current agent position. [(x, y, z)]
-	float disp[3];
-	float dvel[3];		///< The desired velocity of the agent. [(x, y, z)]
-	float nvel[3];
+	float disp[3];          ///< "disp" is just something used during collision detection
```
grahamboree: This could be better described.  We should indicate that it's used only in collision detection in DetourCrowd::update, but also indicate what it actually means in that context, rather than saying it's "just something".

DetourCrowd/Include/DetourCrowd.h
```diff
@@ -143,9 +143,9 @@ struct dtCrowdAgent
...
+	float dvel[3];		///< The desired velocity of the agent. [(x, y, z)] (based on the current path, calculated each frame "from scratch")
+	float nvel[3];          ///< Desired velocity adjusted by obstacle avoidance, calculated each frame "from scratch"
```
grahamboree: I'm not sure we need the `"from scratch"` part in either of these.  Saying something is calculated each frame is sufficient in communicating the value's source.

Kromster80:
This is quote from Miko. Please fell free to cut it

Kromster80:
Although for me the "from scratch" part provides better understanding of what is happening. Meaning it is not just calculated (e.g. relying on previous values), but is actually anew.

grahamboree:
Fair point.  I'll concede the "from scratch" part. :smile: 

I know these were direct quotes from @memononen on the mailing list, but I feel it's important that documentation is more precise and explicit than colloquialisms.  I feel that a something along the lines of  "a temporary value used to accumulate agent displacement during iterative collision resolution" would be much more informative to the reader.

Kromster80:
Please change the comments as they should be.
Back in the days of porting RN to Delphi, those comments helped me a lot, so I have included them here. In other words - better these than none, but if you have even better, then roll them in!

grahamboree:
Okay sounds good.  Thanks for the PR.  
Closing this as I can't modify it.  New changes in #134.

PR #111, opened 2015-09-01
Expression always evaluated to 0

Kromster80:
As far as I know [anything mod 1 = 0]. Please verify and or reject this pull request.

### DIFF
--- a/DetourCrowd/Source/DetourObstacleAvoidance.cpp
+++ b/DetourCrowd/Source/DetourObstacleAvoidance.cpp
@@ -547,8 +547,8 @@ int dtObstacleAvoidanceQuery::sampleVelocityAdaptive(const float* pos, const flo
 	for (int j = 0; j < nr; ++j)
 	{
 		const float r = (float)(nr-j)/(float)nr;
-		pat[npat*2+0] = ddir[0] * r;
-		pat[npat*2+1] = ddir[2] * r;
+		pat[npat*2+0] = ddir[(j%2)*3] * r;
+		pat[npat*2+1] = ddir[(j%2)*3+2] * r;
 		float* last1 = pat + npat*2;
 		float* last2 = last1;
 		npat++;
--- a/DetourCrowd/Source/DetourObstacleAvoidance.cpp
+++ b/DetourCrowd/Source/DetourObstacleAvoidance.cpp
@@ -547,8 +547,8 @@ int dtObstacleAvoidanceQuery::sampleVelocityAdaptive(const float* pos, const flo
 	for (int j = 0; j < nr; ++j)
 	{
 		const float r = (float)(nr-j)/(float)nr;
-		pat[npat*2+0] = ddir[(j%1)*3] * r;
-		pat[npat*2+1] = ddir[(j%1)*3+2] * r;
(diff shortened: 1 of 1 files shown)


wanhconyan:
why i always get this email

发自我的 iPhone

Kromster80:
@wanhconyan: Because you have subscribed to receive these emails from memononen/recastnavigation (Notifications: Watching (be notified of all conversations))

Janiels:
I don't think this is the correct fix. The array has 2 vectors, and making this 0 only ever uses the first. It should probably be `%2` instead of `%1`.
@hymerman @grahamboree Do any of you know the appropriate fix? I have to admit that I do not quite understand the adaptive sampling algorithm (it's outlined in [this](http://digestingduck.blogspot.dk/2010/04/adaptive-rvo-sampling.html) blog post).

memononen:
If should be %2 or &1.

The code in question used to call sin/cos for every sample, and every other ring of samples was offset by half the angle between the samples to cover the are better. A bit hard to example, by say, we have 6 samples on a ring, angle between samples is 360/6 = 60, then odd rings would be offset by 30 degrees. There was a refactoring do to this code to avoid calling sin/cos for each sample, which improved the code quite a bit, but the off looking loop.

The adaptive sampling works so that we take a set of samples on each iteration, pick best sample, shrink the sampling radius and do it again. The goo thing about that sampling is that we get pretty good guess by using less samples, the bad thing is that values that are being sampled are very discontinuous, so sometimes you get temporal flickering in the solution.

Janiels:
Thanks Mikko, I will get that checked in. I didn't know if I should mention you on these issues or not - which would you prefer?

memononen:
I tend to read these as they pop into my inbox and comment as I get the change or have time. If you're stuck, or need my input specifically, please add a mention.

Kromster80:
Glad to see Issues being processed now! :-)

Janiels:
Seein as you're still active @Kromster80, can you upgrade your PR with the changes?

Kromster80:
Done

Janiels:
Thanks! Can you squash your commits so they are ready for the merge?

Kromster80:
How can I do that through github gui and is it really necessary?

Janiels:
It's not necessary, but it helps maintain a clean history in the project. Instead of having these two commits (where the first one is actually completely undone by the second one), we want just the last commit in the master branch.

I don't believe the Github UI can do squashes, but I believe it can open a terminal window for you? If it can, you can use a process like this: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

Kromster80:
Too bad :-(
Sorry, I don't fancy Git (it is over-complicated for my taste) and I hate commandline, so if there's no GUI version of it I won't be able to do it. Sorry.

Janiels:
Another way would be to simply create a new branch and apply the fix there. If you don't want to spend anymore time on it, I understand. I will just open a new PR for it then. Still, thanks for bringing it to our attention!

Kromster80:
Sounds like a plan =)
(not to mention we spent x10 more time on it just commenting)
I'll do the PR now

Janiels:
New PR is #132

PR #119, opened 2015-11-15
Include hit edge index in dtRaycastHit.

richard-fine:
This makes it simpler to investigate exactly what is blocking a ray - for example, if an agent cannot move onto a polygon due to QueryFilter flags, then this provides a way for the agent to find out which flags it needs to have and to perform a state change which alters its movement flags.


memononen:
Thanks for the improvement!

PR #121, opened 2015-11-15
Extract distance field computation from rcErodeWalkableArea

richard-fine:
Pull it out into a separate function to make it easier for people to do custom erosion behaviour based on the intersection of multiple distance fields (e.g. correct handling of 'crawlspace' areas when eroding)

### DIFF
--- a/Recast/Source/RecastArea.cpp
+++ b/Recast/Source/RecastArea.cpp
@@ -26,42 +26,25 @@
 #include "RecastAlloc.h"
 #include "RecastAssert.h"
 
-/// @par 
-/// 
-/// Basically, any spans that are closer to a boundary or obstruction than the specified radius 
-/// are marked as unwalkable.
-///
-/// This method is usually called immediately after the heightfield has been built.
-///
-/// @see rcCompactHeightfield, rcBuildCompactHeightfield, rcConfig::walkableRadius
-bool rcErodeWalkableArea(rcContext* ctx, int radius, rcCompactHeightfield& chf)
+void rcComputeDistanceField(rcCompactHeightfield& chf, unsigned char* dist, int fromAreaID)
 {
-	rcAssert(ctx);
-	
+	rcAssert(dist);
+
 	const int w = chf.width;
 	const int h = chf.height;
-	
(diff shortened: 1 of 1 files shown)


Recast/Source/RecastArea.cpp
```diff
@@ -75,8 +58,8 @@ bool rcErodeWalkableArea(rcContext* ctx, int radius, rcCompactHeightfield& chf)
 						{
 							const int nx = x + rcGetDirOffsetX(dir);
 							const int ny = y + rcGetDirOffsetY(dir);
-							const int nidx = (int)chf.cells[nx+ny*w].index + rcGetCon(s, dir);
-							if (chf.areas[nidx] != RC_NULL_AREA)
+							const int nidx = (int)chf.cells[nx + ny*w].index + rcGetCon(s, dir);
+							if (chf.areas[nidx] != fromAreaID)
```
TheMartin: I don't think this part is entirely correct. This will set the distance of a span to 0 not only when it has at least one neighbor with `fromAreaID`, but also when at least one of the neighbors is missing (`rcGetCon` returns `RC_NOT_CONNECTED`). That makes sense in the original use-case ("expanding" the unwalkable area), but isn't what we want when computing a general distance field.

richard-fine: Now that you mention that, I'm not sure how this code does not potentially end up indexing invalid area indices, because RC_NOT_CONNECTED is defined as 0x3f, so I would have expected us to end up with invalid high nidx values. Are you sure that rcGetCon() is actually capable of returning RC_NOT_CONNECTED in this situation?

TheMartin: Well, no, this block is only entered if `rcGetCon` does _not_ return `RC_NOT_CONNECTED` (see line 57 in your commit), so `nidx` is safe to use. However, the variable `nc` can only be incremented in this block; in other words, the only way it can equal 4 on line 69 is if all four neighbors are connected.

richard-fine: D'oh, now I get you. I've changed it to only do this missing-neighbour check if the field is being based on `RC_NULL_AREA`.

PR #132, opened 2015-12-16
Fix as discussed in PR#111

Kromster80:
Typo fix


Janiels:
The build failure looks unrelated.
@hymerman @grahamboree I don't have rights to label or merge yet, so could one of you merge?

hymerman:
Weird, I'll look into that - I added you as member (since 'owner' has lots of scary buttons like 'delete organisation') but I thought you'd still be able to merge and label.

The build failure is unrelated, and a bit annoying since I stayed up late getting the premake download to work!

Janiels:
I think you have to create a team with rights to this repository, and add me to it explicitly.

hymerman:
Done! Bit of a pain though! I had already set write access to all org repos for members by default, I'm not sure why that setting isn't taking effect. I've added a team called 'write' that allowed write access to this repo, and added all of us to it. That should work.

Janiels:
Yes, looks like that did the trick!

PR #102, opened 2015-06-23
handle the case when allocSpan fails to allocate memory

rafzi:
see #100

PR #133, opened 2015-12-16
Use intermediate variable for nearest point in findNearestPoly

### LINKED ISSUE #118: findNearestPoly may return wrong results
if you pass in the same position pointer for your 'center' parameter as you do the nearestPt 'out' parameter, you will likely get bad results for nearest polygon because the distance metrics while searching for polygons will change since each iteration where a potential result is found will modify the data pointer to by center as well.

recommend using a temporary inside the function and copying the results at the end to allow the same object to be passed in for both.

    dtStatus findNearestPoly(const float* center, const float* extents,
    							 const dtQueryFilter* filter,
    							 dtPolyRef* nearestRef, float* nearestPt) const;

Janiels:
When the same point is used as center and as nearestPt, the function
would override center and give wrong results.

Fix #118

### DIFF
--- a/Detour/Source/DetourNavMeshQuery.cpp
+++ b/Detour/Source/DetourNavMeshQuery.cpp
@@ -721,8 +721,13 @@ dtStatus dtNavMeshQuery::findNearestPoly(const float* center, const float* exten
 	if (dtStatusFailed(queryPolygons(center, extents, filter, polys, &polyCount, MAX_SEARCH)))
 		return DT_FAILURE | DT_INVALID_PARAM;
 	
+	if (polyCount == 0)
+		return DT_SUCCESS;
+	
 	// Find nearest polygon amongst the nearby polygons.
 	dtPolyRef nearest = 0;
+	float nearestPoint[3];
+
 	float nearestDistanceSqr = FLT_MAX;
 	for (int i = 0; i < polyCount; ++i)
 	{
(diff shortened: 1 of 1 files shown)


Janiels:
Actually this is not ready yet. When no polys are found in `queryPolygons`, this will initialize `nearestPt` with uninitialized values.

Janiels:
It should be ready now. I also removed the check on `nearestRef` as it has already been dereferenced earlier in the function.

Detour/Source/DetourNavMeshQuery.cpp
```diff
@@ -721,8 +721,13 @@ dtStatus dtNavMeshQuery::findNearestPoly(const float* center, const float* exten
 	if (dtStatusFailed(queryPolygons(center, extents, filter, polys, &polyCount, MAX_SEARCH)))
 		return DT_FAILURE | DT_INVALID_PARAM;
 	
+	if (polyCount == 0)
+		return DT_SUCCESS;
```
hymerman: Slight behaviour change here; previously `nearestRef` would have been set to 0 if `queryPolygons` returned no polys. We should probably keep that behaviour.

Detour/Source/DetourNavMeshQuery.cpp
```diff
@@ -751,15 +756,17 @@ dtStatus dtNavMeshQuery::findNearestPoly(const float* center, const float* exten
...
+			dtVcopy(nearestPoint, closestPtPoly);
+
 			nearestDistanceSqr = d;
 			nearest = ref;
 		}
 	}
 	
-	if (nearestRef)
-		*nearestRef = nearest;
+	*nearestRef = nearest;
```
hymerman: We should still have the `nullptr` check here... or not, depending on what the outcome of #102 is!

Detour/Source/DetourNavMeshQuery.cpp
Janiels: That's still the behavior - `nearestRef` is assigned to 0 in the beginning of the function.

hymerman: Oh whoops, sorry! I'm imagining PR #102 already being merged in :) Ok; the combination of this PR and #102 should ensure that `*nearestRef = 0` in this case.

Detour/Source/DetourNavMeshQuery.cpp
Janiels: I assume you mean #108. I can make that change in a different PR, when we decide.

hymerman: Whoops :) Of course - sorry, I'm all over the place today!

hymerman:
I'm happy with this as-is; I'll merge it in now since it looks like #108 which would conflict isn't going to be needed

PR #136, opened 2015-12-17
findNearestPoly now fails when nearestRef is null, and only assigns it when the function succeeds

Janiels:
As discussed in #108.


hymerman:
Looks good to me!

Janiels:
Thanks!

Issue #127, opened 2015-12-09
Unit Testing


grahamboree:
We've talked about adding unit tests to some core components of the project.  I thought I'd open a thread to discuss what areas should be prioritized for unit testing and what framework we should use.

**Which framework should we use?**
Personally, my experiences unit testing C++ haven't been as smooth as in other environments, though it's certainly better than nothing.  I've enjoyed using [Catch](https://github.com/philsquared/Catch) for other projects, as it's lightweight, straightforward, and easy to integrate and work with.  [Google Test](https://github.com/google/googletest) is another reasonable alternative, but it's probably a bit too heavy handed for what we'd need.  There are tons of other options as well.  Wikipedia has [a huge list of C++ unit testing frameworks](https://en.wikipedia.org/wiki/List_of_unit_testing_frameworks#C.2B.2B).

**Where do the tests live?**
The tests should not interfere at all with the actual library code or the current integration process.  This probably means that they should probably be in a location that is completely separate from the codebase so that integration doesn't involve a "remove the tests" step.

The tests should be easy to write and run.  This is important in encouraging test creation and maintenance.  If the unit test suite is hard to run, they'll be ignored.  This assertion is subservient to the current codebase's integration experience.  When in doubt, it's better to make the unit testing experience more tedious if in doing so, the experience of using using recast is maintained or improved.

**What should we be testing (first)?**

A few parts of the codebase are pretty straightforward to test, and would benefit from knowing they're a solid foundation.  In particular the majority of geometry and math helper functions are already pure, making testing them trivial.  e.g. `distancePtSeg`, `dividePoly`, etc...

There are also a lot of small helper functions like `insertSort`that are pretty easy to test.

Beyond that, it starts getting into low-level recast functions, which are definitely testable, but might require a bit more domain-knowledge.

grahamboree:
Another aspect to this we should consider: Once we have some unit tests, we should get them outputting gcov information when the game is built by the CI servers.  This will allow us to use things like coveralls to measure code coverage.  

[Here's](https://github.com/nlohmann/json) an example of a C++ project that has CI with Travis and Appveyor set up, along with unit tests using Catch and code coverage with coveralls.  Their unit tests are all in one giant 10k line cpp file, which is less than ideal for sure, but is a good reference point when discussing how to structure our changes.

hymerman:
Haha, we're going to get along well Graham :) This is something I was considering doing already.

[...]

PR #139, opened 2015-12-29
Correctly remove off mesh links to other layers

### LINKED ISSUE #113: Off-mesh connection links not removed between layers in the same tile location
I've been running into some crashes during path finding and after some digging I found some of my tiles ended up with invalid links. This happens when neighboring tiles in the same layer were rebuild, but only for tiles with off-mesh connections between them.

Looking in unconnectExtLinks I found the following check:
```
			if (tile->links[j].side != 0xff &&
				decodePolyIdTile(tile->links[j].ref) == targetNum)
```

side only seems to be set to 0xff when off-mesh connections are created between tiles in the same layer.

It seems neighboring tiles in the same layer are skipped when removing links. Is there any reason for doing so? I removed the check from my code and don't see any side effects.

Janiels:
(As discussed in #113)

External links representing off mesh connections between layers in the
same tile location would not be correctly removed. These links had
side = 0xff which were skipped during tile removal under the assumption
that these links were internal.

Also correct some comments in the same place.

Fix #113

### DIFF
--- a/Detour/Include/DetourNavMesh.h
+++ b/Detour/Include/DetourNavMesh.h
@@ -620,7 +620,7 @@ private:
 	void connectExtOffMeshLinks(dtMeshTile* tile, dtMeshTile* target, int side);
 	
 	/// Removes external links at specified side.
-	void unconnectExtLinks(dtMeshTile* tile, dtMeshTile* target);
+	void unconnectLinks(dtMeshTile* tile, dtMeshTile* target);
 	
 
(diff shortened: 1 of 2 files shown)


hymerman:
I can't say whether the fix looks good or not, but it was given the OK by Mikko so I'll happily accept it :)

Janiels:
Thanks. :+1:
Ideally I wanted to write a regression test for it, but I don't really have any experience with unit testing C++ and I noticed that #127 is still open. I guess we can leave that for the future. :smiley:

PR #141, opened 2016-01-02
Fix heap corruption when collecting region layers

### LINKED ISSUE #30: Found a heap corrupting bug in RecastLayers.cpp
Hey,

I had some hours of fun right now because of a heap corruption. Those are always a bitch to track, but I managed with DUMA ( duma.sourceforge.net/ ).

I noticed the error in these lines in RecastLayers.cpp (lines 315 - 322):
```
                        const int nneis = (int)reg.nneis;
			{
				const unsigned char nei = reg.neis[j];
				rcLayerRegion& regn = regs[nei];
				// Skip already visited.
				if (regn.layerId != 0xff)
					continue;
```

The problem is that nneis can happen to be larger than RC_MAX_NEIS. In my case, nneis was 21 (while the default max is 16), so it was crashing when j became 17 (funny enough, j=16 worked, pure luck IMO).

My fix was to simply increase RC_MAX_NEIS to 32, and it works for now, but that likely does not solve the problem. 

No idea how, where or why nneis is set to a number higher than RC_MAX_NEIS, but I just don't have the time to dig into this right now. Just wanted to let you know.

Janiels:
This fix is the fix Mikko suggested in #30. However, I don't completely understand
the code here. As far as I can see, this code tries to identify which regions should be merged into what will finally become the layers. @memononen, can you try explaining why it's ok that we do not collect all of these regions? Would it not be better to expand the array to be able to collect all the neighbours? Or will this situation just introduce a few more layers, which is probably ok?

The collecting could overflow the neighbours array due to a missing boundary
check.

Fix #30

### DIFF
--- a/Recast/Source/RecastLayers.cpp
+++ b/Recast/Source/RecastLayers.cpp
@@ -258,7 +258,7 @@ bool rcBuildHeightfieldLayers(rcContext* ctx, rcCompactHeightfield& chf,
 						const int ay = y + rcGetDirOffsetY(dir);
 						const int ai = (int)chf.cells[ax+ay*w].index + rcGetCon(s, dir);
 						const unsigned char rai = srcReg[ai];
-						if (rai != 0xff && rai != ri)
+						if (rai != 0xff && rai != ri && regs[ri].nneis < RC_MAX_NEIS)
 							addUnique(regs[ri].neis, regs[ri].nneis, rai);
 					}
 				}


memononen:
`regs[ri].neis` can only contain up to `RC_MAX_NEIS` items, the add unique does not check for overflow.
Maybe it would be cleaner to pass `RC_MAX_NEIS` to `addUnique` instead and have the overflow check there. The reason I had the overflow check as a separate if was that it should make it more explicit that it was checking for overflow.

Janiels:
That makes sense. However, by just checking for overflow (and not expanding the list, for example by increasing `RC_MAX_NEIS` to `32`), `rai` is not added to the list. Will this not cause any problems?

memononen:
Not adding `rai` will prevent regions to merge. It should not break anything, but you'll potentially get more regions.

Janiels:
Ok, that sounds fine.

GregSlazinski:
Thank you for clarification, I see Mikko's comment in #141 

```
What I mean is that silently ignoring the layer will probably produce a wrong mesh (i.e. by not adding it). Instead the function should fail early when it detects that there will be too many layers.

This code has been in R&D for a long time so at this point I am confident that the 63 layer count is enough that it never overflows. However, it would still be good to be safe about it.
```

However as for doing the boundary safety checks, I'm quite certain they should be performed always, without doing any assumptions.
Even if you get overflow, it doesn't mean the app will crash, instead the app may crash, or you might as well corrupt some other unrelated memory.
Overflowing buffer only corrupts the stack memory which will result in undefined behavior.

Janiels:
> However as for doing the boundary safety checks, I'm quite certain they should be performed always, without doing any assumptions.

I agree, but it should be detected and handled in `rcBuildHeightfieldLayers` -- not just `addUnique`.

Janiels:
Done in #214.

PR #145, opened 2016-01-05
Ensure SDL2.dll is copied for paths with spaces

Janiels:
Just a small fix for when the solution/bin path contains spaces. This was the only thing
preventing me from running it directly after premake generation.


grahamboree:
Cool, thanks!  Will this cause issues with paths without spaces?  I don't know much cmd.exe...

grahamboree:
Cool, just tested it and it's working fine.  LGTM

PR #147, opened 2016-01-06
Unit testing framework.

### LINKED ISSUE #127: Unit Testing

grahamboree:
I took a stab at setting up unit testing for the project using [Catch](https://github.com/philsquared/Catch).

It's all contained in a folder called "Tests" at the root.  I figured it'd be nice to split up the unit tests by which component and which file they correspond to.  For example, I added a simple test for `rcVdot`, a function that's implemented in `Recast/Recast.h`.  I created a corresponding test file `Tests/Recast/Tests_Recast.cpp` for this unit test.  My thinking about this is that we want to easily be able to find unit tests that correspond to a given file, and also not to shadow filenames of library files (hence the "Tests_" prefix).

There's a new project in premake to create an executable that runs all the unit tests.  While the "Tests" project doesn't need most of the resources it includes and links against, I wanted to try to ensure that the build configuration would work if/when we start adding tests for components that require 3rd party libraries.

I also set up Travis and Appveyor to build and run the tests with every checkin.  If a unit test fails, the build fails.

I think this is a pretty reasonable start, and similar to other similar open source projects I looked at.  I'm open to suggestions if people have of ways of improving it.

Commit description:
* Tests live in a "Tests" folder alongside the other components.  Inside the "Tests" folder, tests are split into folders by the component they test.  For example, the example unit test of "rcVdot" (which is implmemented in Recast/Recast.h) lives in "Tests/Recast/Tests_Recast.h".
* Uses the Catch testing framework
* One example test of "rcVdot"
* Tests are run on Travis and Appveyor after every build.  Failing unit tests will fail the build in both case.
* Added instructions on running unit tests to the readme

Fixes #127


hymerman:
A fantastic start - thanks Graham! I'm happy to merge this in as it is.

Issue #146, opened 2016-01-05
Wrong detail mesh sampling with layer partitioning


Janiels:
The following case happens in the default `nav_test.obj` with `Max Climb = 1.6` and `Partitioning = Layers`:
![bad sampling](https://cloud.githubusercontent.com/assets/7887810/12126113/9234b1ac-b3ef-11e5-8ba6-1148e4a8472d.png)
This only affects the detail mesh; the regular poly mesh does not have the problem.
However, the problem originates in `rcBuildPolyMesh`, not `rcBuildPolyMeshDetail`. These
are the regions created beforehand:
![bad sampling regions](https://cloud.githubusercontent.com/assets/7887810/12126256/5ec7f030-b3f0-11e5-818b-c1abc0045724.png)
And this is the poly mesh:
![bad sampling polys](https://cloud.githubusercontent.com/assets/7887810/12126278/85e9f19a-b3f0-11e5-9f34-71f9a105e329.png)
The problem is that all polys in the poly mesh have the same region ID, even though the
compact heightfield has two different regions here. This causes the detail sampling
to sample the wrong height.

It appears the region ID of the stair-poly disappears when one of the border vertices are
removed during the build of the poly mesh. The vertex is removed, and when the hole
created is later converted to a polygon again, the wrong region ID is used.
I haven't been able to debug it any further - I got kind of stuck in the hole triangulation. :smile:

Janiels:
I took another look. These are the vertices being removed (well, first merged, then removed) when the poly mesh is built:
![bad sampling vertices removed](https://cloud.githubusercontent.com/assets/7887810/12187355/723fe1a6-b5ab-11e5-8b90-563677227fe7.png)
Without removing those vertices, the poly mesh is the following:
![bad sampling poly after removal](https://cloud.githubusercontent.com/assets/7887810/12187399/e56e8f9c-b5ab-11e5-9a9d-a664cea65f42.png)
I have marked the polygon that is created by removing those circled vertices as well. We end up giving the region ID of the stair-contour to only vertex 3 and 4 (vertex 2 is shared and receives the region ID of the ground contour). In the end this means that when the polygon is triangulated, all 3 triangles created for this poly have the same polygon ID, and these are used as the initial polys.

PR #148, opened 2016-01-08
Improve area and region ID selection for triangulated holes

Janiels:
This improves on a case where Recast would assign a region and area ID
that did not end up matching the final poly mesh very well.
This would happen in cases where several vertices were shared by
polygons with different regions after removal of a border vertex.
This fix improves the case by selecting the most common region and
area ID instead of always using the first; this should resolve the
majority of these cases. See #146 for more detailed information.

This does not catch all situations; it's possible that vertex 3 (in the picture in the second post of #146) would be shared with a different polygon. In this case it would be possible that there was no common region/area ID. However, this fix should fix the majority of (already rare) cases.
A better solution would need to take into account the shared vertices and all their possible types. Maybe @memononen has some thoughts on a better solution?

### DIFF
--- a/Recast/Source/RecastMesh.cpp
+++ b/Recast/Source/RecastMesh.cpp
@@ -663,6 +663,16 @@ static bool canRemoveVertex(rcContext* ctx, rcPolyMesh& mesh, const unsigned sho
 	return true;
 }
 
+static int getCommonOrFirst(int val1, int val2, int val3)
+{
+	if (val1 == val2 || val1 == val3)
+		return val1;
+	if (val2 == val3)
+		return val2;
+
+	return val1;
+}
+
 static bool removeVertex(rcContext* ctx, rcPolyMesh& mesh, const unsigned short rem, const int maxTris)
 {
 	const int nvp = mesh.nvp;
@@ -895,8 +905,8 @@ static bool removeVertex(rcContext* ctx, rcPolyMesh& mesh, const unsigned short 
 			polys[npolys*nvp+0] = (unsigned short)hole[t[0]];
 			polys[npolys*nvp+1] = (unsigned short)hole[t[1]];
 			polys[npolys*nvp+2] = (unsigned short)hole[t[2]];
-			pregs[npolys] = (unsigned short)hreg[t[0]];
-			pareas[npolys] = (unsigned char)harea[t[0]];
+			pregs[npolys] = (unsigned short)getCommonOrFirst(hreg[t[0]], hreg[t[1]], hreg[t[2]]);
+			pareas[npolys] = (unsigned char)getCommonOrFirst(harea[t[0]], harea[t[1]], harea[t[2]]);
 			npolys++;
 		}
 	}


Janiels:
I'm not actually sure that the situation I mentioned can exist; the polygon touching vertex 3 would also have to touch the vertex being deleted. I will try to run this change on a large map and see if I can find a case.

Janiels:
Here's a case where the fix in this PR will not do improve the region/area selected. These happen with watershed partitioning. Contours:
case contours
Final poly mesh detail:
case polymesh

The situation happens after removal of the first few border vertices. The polymesh looks like this and the circled vertex is about to be removed:
case polymesh
One of the triangles/initial polygons ends up being this one, has three different region IDs:
case polymesh


These cases do not cause any problems with the sampling as the regions do not overlap in the XZ plane. It's still a little alarming to see all these regions disappear though, and the mesh isn't the best.  This 'problem' is actually what `Max Edge Length` is meant to solve, as far as I understand.

It's also pretty easy to see that it's impossible to pick a "correct" region ID in many cases. Perhaps a better solution would be to get rid of regions entirely during the build of the poly mesh and do the detail sampling in a different way. The regions simply cannot be reliable after vertices have been removed.

memononen:
The `Max Edge Length` only applies to non-tile edges. They get extra subdivisions which improves the tessellation a lot of various cases. There are many cases where this would lead better triangulation tile edges too, but I chose try to keep tile edges straight. Detour can connect non-matching subdivision on tile edges, though.

I think one option would be to mark polygons which are merged and use the flood fill fallback in the detail mesh build phase. Matching polygons with the heightfield seemed initially like a simple thing, but ended up being very brittle.

Janiels:
So you mean mark the polygons with the regions of all the polygons they were created from?
That would allow us to remove the flood fill too, I believe. Which might actually cause a significant performance gain.

Janiels:
Oh, I think I understand what you mean now. Skip the entire first part of `getHeightData` if the polygon has been merged due to vertex removal, since it could potentially have this overlap problem.
I'm curious, what is the point of the first part of `getHeightData`? It looks like the flood fill alone should be able to collect all height data required.

I think there is still a potential problem in `getHeightDataSeedsFromVertices`, though I'm not sure how often if ever it can happen. I imagine this situation could happen due to the DFS search for the center:
dfs
The AABB of the polygon being sampled would have to be slightly larger than in the picture shown, but I think that is possible (eg. just imagine it rotated 45 degrees).
What do you think? Since the polys are convex it would be relatively simple to make a mini-A*. Or BFS would probably fix it too.

Janiels:
I will create a new PR instead.

Janiels:
New PR is at #151.

memononen:
Yes, you got the idea just right :)

I think I may have thought I was using breath-first in the current version. In that case the highlighted route would be longer than from one of the other vertices. It is possible that there is no line of sight from a vertex to the center of the polygon, due to quantization and all that.

I think the code in the PR looks good. A more paranoid version would always add all the 4 neighbours, but visit the one directing towards the goal first.

PR #150, opened 2016-01-10
Fix time measuring on Windows

Janiels:
The demo relied on the build context returning accumulated times in
microseconds. On Windows this was using QueryPerformanceCounter directly
which has a higher resolution, thus returning wrong measured times. We
now report the accumulated times in microseconds on all platforms.

### DIFF
--- a/RecastDemo/Include/PerfTimer.h
+++ b/RecastDemo/Include/PerfTimer.h
@@ -27,6 +27,6 @@ typedef __int64 TimeVal;
 #endif
 
 TimeVal getPerfTime();
-int getPerfDeltaTimeUsec(const TimeVal start, const TimeVal end);
+int getPerfTimeUsec(const TimeVal duration);
 
 #endif // PERFTIMER_H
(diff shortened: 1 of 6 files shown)


hymerman:
LGTM.

Note: I tend to freak out about times being represented in 32 bit integers, which they get truncated to at some points in this implementation of profiling for the demo. Since they're counts of microseconds they wrap after 30 minutes or so, which is admittedly a very long time for a Recast build but not impossible (on Alien Isolation's largest levels, with lots of custom stuff going on, before optimisation, some builds were ~40 mins). We should change that at some point! That's an existing problem though; this PR is a strict improvement over what's there so I'm happy for it to be merged.

Janiels:
That did irk me too (that is why I changed `m_accTime` to `TimeVal`). But changing `doGetAccumulatedTime` is breaking, unfortunately. :(

hymerman:
Yeah, that's better :) Ok, we'll leave that kind of thing alone for now. I'll merge this now.

PR #151, opened 2016-01-12
Use flood fill for potentially overlapping polys

Janiels:
Use flood fill for potentially overlapping polys

This changes the detail sampling to find height data by using the flood
fill method for polygons that were merged because of a border vertex
removal. This could potentially cause overlapping polygons to receive
the same region IDs which would later make the fast-path sampling in
getHeightData sample the wrong heights.

Also includes optimizations for this path and getHeightData itself.
Previously the seeding was responsible for 25% of the total time spent
in getHeightData in these cases. We now use the direction and prune the
nodes when we move towards the center cell, instead of doing a
full-blown DFS. This makes the overhead of seedArrayWithPolyCenter
virtually nonexistent (~2% according to profiling).

In my tests this new path was only reached for about 4% of polygons
(although that was with a very large tile size, for smaller tile sizes
it will probably be larger). Arguably the optimization is not necessary,
but the code is also a fair bit shorter and maybe a bit easier to
understand. I ran the new center-finding-algorithm on a huge map with
around 3.2 million polygons/heightfields and in no cases did it fail to
find the center.

I have also optimized the inner loop of `getHeightData`. It turns out
the 3x `rcIntArray::push` were accounting for around 11% of time spent
in this method, with 5% being function call overhead to
`rcIntArray::resize` (even without doing any resizing - pure function
call overhead!), so I changed this slightly to reduce code bloat. I also
changed `rcIntArray::resize` so the common case can be inlined, since
`rcIntArray::push` is used from all over the code base. Finally I
changed the lookup tables in `rcGetDirOffsetX` and `rcGetDirOffsetY` to
be static. MSVC generates much better code for them when they are
static.

### DIFF
--- a/Recast/Include/Recast.h
+++ b/Recast/Include/Recast.h
@@ -513,6 +513,14 @@ void rcFreePolyMeshDetail(rcPolyMeshDetail* dmesh);
 /// @see rcCompactSpan::reg
 static const unsigned short RC_BORDER_REG = 0x8000;
 
+/// Polygon touches multiple regions.
+/// If a polygon has this region ID it was merged with or created
+/// from polygons of different regions during the polymesh
+/// build step that removes redundant border vertices. 
+/// (Used during the polymesh and detail polymesh build processes)
+/// @see rcPolyMesh::regs
+static const unsigned short RC_MULTIPLE_REGS = 0;
+
 /// Border vertex flag.
 /// If a region ID has this bit set, then the associated element lies on
 /// a tile border. If a contour vertex's region ID has this bit set, the 
@@ -1059,7 +1067,7 @@ inline int rcGetCon(const rcCompactSpan& s, int dir)
 ///  	in the direction.
 inline int rcGetDirOffsetX(int dir)
 {
-	const int offset[4] = { -1, 0, 1, 0, };
+	static const int offset[4] = { -1, 0, 1, 0, };
 	return offset[dir&0x03];
 }
 
@@ -1069,8 +1077,18 @@ inline int rcGetDirOffsetX(int dir)
 ///  	in the direction.
 inline int rcGetDirOffsetY(int dir)
 {
-	const int offset[4] = { 0, 1, 0, -1 };
+	static const int offset[4] = { 0, 1, 0, -1 };
 	return offset[dir&0x03];
+}
+
+/// Gets the direction for the specified offset. One of x and y should be 0.
+///  @param[in]		x		The x offset. [Limits: -1 <= value <= 1]
+///  @param[in]		y		The y offset. [Limits: -1 <= value <= 1]
+///  @return The direction that represents the offset.
+inline int rcGetDirForOffset(int x, int y)
+{
+	static const int dirs[5] = { 3, 0, -1, 2, 1 };
+	return dirs[((y+1)<<1)+x];
 }
 
 /// @}
--- a/Recast/Include/RecastAlloc.h
+++ b/Recast/Include/RecastAlloc.h
@@ -64,40 +64,54 @@ class rcIntArray
 	int m_size, m_cap;
 	inline rcIntArray(const rcIntArray&);
 	inline rcIntArray& operator=(const rcIntArray&);
+
+	void doResize(int n);
 public:
 
 	/// Constructs an instance with an initial array size of zero.
-	inline rcIntArray() : m_data(0), m_size(0), m_cap(0) {}
+	rcIntArray() : m_data(0), m_size(0), m_cap(0) {}
 
 	/// Constructs an instance initialized to the specified size.
 	///  @param[in]		n	The initial size of the integer array.
-	inline rcIntArray(int n) : m_data(0), m_size(0), m_cap(0) { resize(n); }
-	inline ~rcIntArray() { rcFree(m_data); }
+	rcIntArray(int n) : m_data(0), m_size(0), m_cap(0) { resize(n); }
+	~rcIntArray() { rcFree(m_data); }
 
 	/// Specifies the new size of the integer array.
 	///  @param[in]		n	The new size of the integer array.
-	void resize(int n);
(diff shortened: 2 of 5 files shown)


Janiels:
In #148:
> I think the code in the PR looks good. A more paranoid version would always add all the 4 neighbours, but visit the one directing towards the goal first.

I will make that change. I think it worked reliably in my tests due to heightfield erosion - once you remove that, I have a feeling the 2-directions only would start breaking down. The detail sampling is still finicky when the heightfield is not eroded - it creates a lot of triangles extending far upwards - but that does not mean I should break it even more.

Janiels:
Ok, new changes are up. Will now move in all 4 directions, but prefer one of the directions that move us towards the center.

memononen:
I think those spikes are caused by missing data (`getHeight` fails to find anything sane), or get height just returning data from a floor below/above. One option would be to constraint the flood fill to be inside the polygon, plus few voxels so that we capture thin polys too, and dilate the rest.

Janiels:
I think so too:
detail sampling
CHF:
compact heightfield
Polygon, which after simplification goes through the obstacle:
simplified polygon
So the heightpatch has no heights here (i.e. they are `RC_UNSET_HEIGHT = 0xffff`). And the detail sampling will prefer those heights since they have very big errors.

Another solution could be to change `getHeight` to look several cells away (we need to look up to `ceil(MaxEdgeError)` cells away, not just `1`, I believe). I don't quite understand what you mean with limiting the flood fill to inside the polygon - would this not give us heights off by a few voxels, in some cases? The current method should give correct heights for those cases, I think.

memononen:
There are two cases where we get bad values. One is the above, where there just is no value. Another one is where we get wrong height because the nearest voxel is downstairs (because of flood filling). Basically the staircase case you had earlier in #146 .

Constraining the search to be inside the polygon would get rid of those down-stairs-samples, but we would have empty samples instead. Dilation or larger search could solve that.

Those are the two cases I've seen in the past. It is very much possible that I have witnessed the problem you described in #146, and the cases are separate.

One option to search further is to use BFS, another is spiral search pattern, like this (from old usenet post, iirc, lost the source):
```
	// Search in a spiral
	const int cx = ...;
	const int cy = ...;
	const int rad = ...;

	int x = 0, y = 0, dx = 0, dy = -1;
	int maxSize = rad*2 + 1;
	int maxIter = maxSize * maxSize;

	for (int i = 0; i < maxIter; i++) {
		const int lx = cx + x;
		const int ly = cy + y;

		// Do stuff (might need to bounds check lx, ly)

		if ((x == y) || ((x < 0) && (x == -y)) || ((x > 0) && (x == 1-y))) {
			int tmp = dx;
			dx = -dy;
			dy = tmp;
		}
		x += dx;
		y += dy;
	}
```

Janiels:
I think the nearest-voxel case should be fixed by this PR, since it should now always find the center of the poly (at least in the very vast majority of cases) due to the direct search. Previously this should only have been a problem if the initial seeding found the wrong center, as far as I can see. The BFS in `getHeightData` should not be able to visit downstairs polys if the seed is placed on the poly, since the poly is convex.

I tried your spiral search in `getHeight` and it fixes the problem when I up the radius to `2`. Do you think passing the max edge error through the build process and using it here would be an ok fix?
And what do you think about the code in this PR?

memononen:
Yeah, maybe something like `max(1, (int)ceil(edgeError))`, should do it. Or just leave it to 2 or 3.

Generally the PR LGTM.

Janiels:
Great. I've opened #153 for the other issue. @hymerman @grahamboree not sure if one of you two want to review this one too?

hymerman:
Code looks ok, I don't have time right now to run it and inspect results though, sorry. Are the odd cases in the images problems that existed before this PR? And are they only for cases where no erosion is done? If the answer to either of those questions is 'yes' then let's merge!

As a side note, I'd love to add some functional tests alongside our unit tests, to build navmeshes from example geometry and compare metrics with previous versions, to flag unintended differences that this kind of changelist has the potential to introduce. You know, "whoops, there are 2x as many polys now" or "total poly area is far less than before for really tall maps", things that would suggest subtle new problems. I'm not sure how feasible that is on Travis or AppVeyor but I'm having a look at doing it locally.

Janiels:
Yes, the cases existed before this PR, and only without erosion. They will by fixed by #154.
I agree some large-scale functional tests would be very nice. Also a set of benchmarks would be good to have and be able to run, to see how changes affected them. My initial version of this PR made detail sampling 5 times slower when I benchmarked it on the data set of the company I work for. So it would be very nice if we could have a huge list of geometries that we could automatically run such changes on. Unfortunately that data set is copy righted so I cannot share it. :cry:
Not sure where we could get such a list of triangle meshes representing real world data. Does anyone have an idea?

hymerman:
Fair enough, let's merge this in then! I'll let you pull the trigger :)

I'll create a new issue for discussing functional tests.

hymerman:
Opened #155 to discuss functional testing.

jackpoz:
> Not sure where we could get such a list of triangle meshes representing real world data. Does anyone have an idea?

https://github.com/TrinityCore/TrinityCore/tree/3.3.5 provides a tool to get around 1.56 GB of Recast data, in particular argument https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/tools/mmaps_generator/PathGenerator.cpp#L182 will generate .obj files of tiles like the ones shown at https://github.com/recastnavigation/recastnavigation/pull/18 . 

I'm attaching an .obj as example, if interested we have around 3500 of them. The only "issue" is that we don't actually own the copyright of the mesh, that's something to check before including anything the repo.

[map5711827.zip](https://github.com/recastnavigation/recastnavigation/files/90981/map5711827.zip)

Janiels:
Yes, I think it's problematic to include those as it looks like that data comes from the WoW files.

PR #152, opened 2016-01-13
Add documentation about data to addTile

### LINKED ISSUE #149: addTile should document explicitly that it needs its own copy of the data passed
When adding a tile to a `NavMesh`, detour will store `data`. However, it is not documented anywhere  that Detour makes changes to this data at runtime (for example changes to `tile->links`, which points into `data`). Even if the parameter is not const, we should document this explicitly. See #138.

Janiels:
dtNavMesh needs exclusive access to the data passed to addTile since it
will make dynamic changes to this at runtime. For that reason the data
should not be reused until the tile has been removed from the nav mesh
again.

Fix #149


grahamboree:
LGTM

PR #154, opened 2016-01-14
Increase recover radius for unsampled heights

### LINKED ISSUE #153: Wrong detail mesh sampling when height field is not eroded
When the height field is not eroded, the recovery case in `getHeight`, which looks 1 cell further out, is often not enough. More details in #151.

Janiels:
When the detail mesh sampling sees an unset height it previously tried
to recover by looking for heights of neighbors. When the heightfield was
not eroded this recovery would fail since the contour simplification
could mean that the nearest valid cell was further away than a neighbor.
We now walk adjacent cells in a spiral to find a height and do this
depending on the max simplification error.

Fix #153

### DIFF
--- a/Recast/Source/RecastMeshDetail.cpp
+++ b/Recast/Source/RecastMeshDetail.cpp
@@ -202,7 +202,7 @@ static float distToPoly(int nvert, const float* verts, const float* p)
 
 static unsigned short getHeight(const float fx, const float fy, const float fz,
 								const float /*cs*/, const float ics, const float ch,
-								const rcHeightPatch& hp)
+								const int radius, const rcHeightPatch& hp)
 {
 	int ix = (int)floorf(fx*ics + 0.01f);
 	int iz = (int)floorf(fz*ics + 0.01f);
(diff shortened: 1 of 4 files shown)


Recast/Source/RecastMeshDetail.cpp
```diff
@@ -1130,6 +1162,7 @@ bool rcBuildPolyMeshDetail(rcContext* ctx, const rcPolyMesh& mesh, const rcCompa
 	const float ch = mesh.ch;
 	const float* orig = mesh.bmin;
 	const int borderSize = mesh.borderSize;
+	const float maxEdgeError = mesh.maxEdgeError;
```
memononen: would make sense to calculate the radius here, instead of in `getHeight()`

Recast/Source/RecastMeshDetail.cpp
```diff
@@ -211,24 +211,56 @@ static unsigned short getHeight(const float fx, const float fy, const float fz,
...
+			}
+
+			// Abort the search if we're entering a new 'ring' and have found a result
+			if (i + 1 == nextRingIterStart)
+			{
+				if (h != RC_UNSET_HEIGHT)
+					break;
+
+				nextRingIterStart += nextRingIters;
+				// Each successive 'ring' has 8 additional cells in it
```
memononen: The ring and 8 does not make sense, can you elaborate?
I think the reference to the spiral loop is from here (I thought it was usenet thing, but nope):
http://stackoverflow.com/questions/398299/looping-in-a-spiral

Janiels: Basically the rings I'm talking about are these:
grid
I don't know a better word for it. Maybe bounds?
The blue 'ring' has 8 cells. The red has 16. The green has 24, and so on. Each time we 'expand' the ring we add 2 cells around the center of each side, so we add 8 cells in total.
I added the check to minimize the impact of this change. Basically if we find a height at `distance = 1`, we stop there already, instead of continuing to outer rings. This means the change will abort here, for the vast majority of cases, and it will be the exact same behavior as before.
The only cases we need to continue the search are the rare ones, where we didn't find a height in the first ring.
Does that make sense? In any case I will add a better comment here.

memononen: That's right! I was naively expecting quadratic expansion, but it is outline, linear :) Ring is good term, just explain why 8 and it's golden.

Janiels:
Feedback addressed, new version.

memononen:
:+1:

Janiels:
Thanks, I've merged it then.

PR #156, opened 2016-01-14
Remove BitDeli badge from README.md

hymerman:
It appears that BitDeli is defunct; their website has had a banner about a change in the GitHub API for a long time, which is a dead link to a blog article, and none of the example stats have been updated for years. Let's just remove it!


Janiels:
LGTM

PR #157, opened 2016-01-14
Fix possible compile/link error with multiple BuildContext definitions

### LINKED ISSUE #44: Compilation error in RecastDemo: multiple BuildContext
multiple definition of `BuildContext::~BuildContext()':

Linking executable: ..\..\Bin\RecastDemo.exe
lib\Debug/libDetourTileCache.a(DetourTileCache.o):DetourTileCache.cpp:(.text$_ZN12BuildContextD1Ev[BuildContext::~BuildContext()]+0x0): multiple definition of `BuildContext::~BuildContext()'
obj\Debug\RecastDemo\Source\SampleInterfaces.o:RecastDemo/Source/SampleInterfaces.cpp:26: first defined here
using master branch.

Indeed BuildContext is declared twice;
- once in the lib DetourCacheTile
- once in SampleInterfaces.cpp/h

Example of solution: rename BuildContext struct in the DetourCacheTile lib by dtcDetourCacheTile or use a namespace.

hymerman:
There is a class and a struct called BuildContext in two different files,
one .h and one .cpp. Depending on how the user structures their program,
it's possible that this would cause a compile or link error, as seems to
have happened in #44. I've just renamed the struct to get around this.

Fixes #44

### DIFF
--- a/DetourTileCache/Source/DetourTileCache.cpp
+++ b/DetourTileCache/Source/DetourTileCache.cpp
@@ -40,10 +40,10 @@ inline int computeTileHash(int x, int y, const int mask)
 }
 
 
-struct BuildContext
+struct NavMeshTileBuildContext
 {
-	inline BuildContext(struct dtTileCacheAlloc* a) : layer(0), lcset(0), lmesh(0), alloc(a) {}
-	inline ~BuildContext() { purge(); }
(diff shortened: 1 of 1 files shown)


Janiels:
LGTM

PR #158, opened 2016-01-15
Use size_t for rcAlloc and dtAlloc

hymerman:
This was already partly done by @cexikitin in #65, I'm just finishing it
off in their absence!

Closes #65


grahamboree:
:+1:

PR #159, opened 2016-01-15
Add Issue Stats badges to README.md

hymerman:
The readme was looking lonely after the BitDeli badge was taken away. Here are some nice badges to replace it, which also helpfully show how active a project this has become! They show recent average time to close issues and PRs.


grahamboree:
:+1:

PR #160, opened 2016-01-17
Add settings storage to geometry sets

Janiels:
This adds the ability for geometry sets to store build settings that can
be automatically applied when they are loaded. This should allow sharing
of .gset files to demonstrate problems with certain settings on certain
files. It also allows people to diagnose problems more easily by being
able to dump their own triangle meshes and settings and load them in the
demo, with all of its visualization options. .gset files can be created
from the current mesh and settings by pressing the 9 key, which will
generate it in the same folder as the input mesh.

Also converts more of the demo to use STL.

The reason I introduced `navMeshBMin` and `navMeshBMax` is because it allows people
to use the demo to diagnose problems more easily. I (and I'm sure Mikko) have noticed a general
trend in visualizers for the meshes to be -- well poor, compared to RecastDemo (and so is mine!).
It should be easier for people to dump their problematic areas to a `.obj` and `.gset`, and include
the areas to mesh with their own settings now.

Additionally I have changed the `.gitignore` to ignore all files in `Meshes/`. I think most files will be there because we/people are testing on other meshes, not because we want to add them to the repo - if we really want to we can always do it with `git add -f`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,8 +16,8 @@ RecastDemo/Bin/Tests
 # Build directory
 RecastDemo/Build
 
-# Ignore some meshes based on name
-RecastDemo/Bin/Meshes/_*
+# Ignore meshes
+RecastDemo/Bin/Meshes/*
 
 ## Logs and databases #
 *.log
--- a/RecastDemo/Include/Filelist.h
+++ b/RecastDemo/Include/Filelist.h
@@ -22,6 +22,7 @@
 #include <vector>
 #include <string>
 
-void scanDirectory(std::string path, std::string ext, std::vector<std::string>& fileList);
+void scanDirectoryAppend(const std::string& path, const std::string& ext, std::vector<std::string>& fileList);
+void scanDirectory(const std::string& path, const std::string& ext, std::vector<std::string>& fileList);
 
 #endif // FILELIST_H
--- a/RecastDemo/Include/InputGeom.h
+++ b/RecastDemo/Include/InputGeom.h
@@ -31,11 +31,51 @@ struct ConvexVolume
 	int area;
 };
 
+struct BuildSettings
+{
+	// Cell size in world units
+	float cellSize;
+	// Cell height in world units
+	float cellHeight;
+	// Agent height in world units
+	float agentHeight;
+	// Agent radius in world units
+	float agentRadius;
+	// Agent max climb in world units
+	float agentMaxClimb;
+	// Agent max slope in degrees
+	float agentMaxSlope;
+	// Region minimum size in voxels.
+	// regionMinSize = sqrt(regionMinArea)
+	float regionMinSize;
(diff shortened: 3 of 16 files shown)


Janiels:
Looks like the build break is because of a `nullptr` usage. Do we want to enable C++11 (or perhaps even C++14) in the demo? It will be inconsistent with other null pointers, so I can change it regardless, but what do you guys think?

grahamboree:
It'd be interesting to see what compilers are popular today among game devs.  Obviously MSVC is up there as it's the primary compiler for xbox and windows.  Clang is probably the second most widely used, as it's part of the OSX, iOS and PS4 toolchains.  GCC is certainly up there as well, since the Android NDK is based on it.
Apart from that I'm not really sure if people are still using weird compilers for console dev.  IIRC PSP used a bizarre version of Intel's compiler at one point, but I'm fairly certain the Vita is using MSVC now.

To me, the stuff we'd like to keep at the lowest reasonably common compiler denominator is what'll be compiled into game runtimes.  As long as there's _a way_ to compile RecastDemo and the associated tooling for each platform, I think it's best that we stay modern.  C++11 and 14 offer a bunch of great features that are designed to improve code legibility and performance, and it'd be great to take advantage of them if we can.

http://en.cppreference.com/w/cpp/compiler_support

C++11 support seems to be pretty pervasive.  The only outliers are older, rarely used platforms (Digital Mars and HP).  C++14 is supported fully by Clang and GCC, and MSVC is only missing some of the more esoteric features.

Janiels:
Are people even compiling and running the demo with those platforms? I completely get it for Recast/Detour itself, I don't think we need to enable C++11/14 for those, but for the demo I doubt there are many, if any, that compile and run it on the more 'exotic' platforms. Besides, SDL 2 only officially supports Windows, OS X, Linux, iOS and Android.

grahamboree:
Totally agree.  I think as long as there's some way for the demo to work on Windows, OSX and Linux with open-source or free resources (VS2k15 community, gcc, clang) any 11/14 features are fine, and should be encouraged.  
This is not the same for R&D where it might be compiled for some weird platform with special requirements.  This is the same argument for not using STL in R&D.

For this PR, I think as long as the appveyor config is set to only build for platforms that support nullptr_t (I think that's 2010+) it should be fine.

grahamboree:
Nevermind, looks like the nullptr's got switched back to 0's.  :stuck_out_tongue: This change looks good as is, so I'll merge it now.  I think it'll be easier to tackle C++11/14 support in another thread, in case there are any issues with the build pipeline.

Janiels:
Thanks. :+1:

[I did read the other day](http://genbattle.bitbucket.org/blog/2016/01/17/c++-travis-ci/) that C++11 support in Travis requires some changes, so it's probably best to do that in a separate PR anyway.

PR #165, opened 2016-01-19
Change if to else-if to avoid unnecessary comparison

wbierman:


wbierman:
I believe the compilation failure listed above with 'AppVeyor' is incorrect.

grahamboree:
Yeah it looks like AppVeyor couldn't spawn a worker to run the VS2010 build, which is why it failed.  I'm not a huge fan of if's and else's without braces; it's inviting a dangling else bug.  Could you add them?

wbierman:
Done.  I replaced the previous commit.  Maybe this time AppVeyor will succeed!

grahamboree:
Thanks!

PR #166, opened 2016-01-19
Fixed rule of 3 violations

grahamboree:
Added virtual qualifier to a few destructors in subclasses.
Removed a few empty destructors that didn't need to be explicitly defined.
Fixed a few typos and some minor code formatting

### DIFF
--- a/DebugUtils/Include/DebugDraw.h
+++ b/DebugUtils/Include/DebugDraw.h
@@ -210,6 +210,10 @@ public:
 	virtual void end();
 	void clear();
 	void draw(struct duDebugDraw* dd);
+private:
+	// Explicitly disabled copy constructor and copy assignment operator.
+	duDisplayList(const duDisplayList&);
+	duDisplayList& operator=(const duDisplayList&);
 };
 
 
--- a/Detour/Include/DetourNavMesh.h
+++ b/Detour/Include/DetourNavMesh.h
@@ -145,7 +145,7 @@ enum dtPolyTypes
 };
 
 
-/// Defines a polyogn within a dtMeshTile object.
+/// Defines a polygon within a dtMeshTile object.
 /// @ingroup detour
 struct dtPoly
 {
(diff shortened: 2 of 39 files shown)


grahamboree:
Okay.....weirdly gcc and MSVC don't agree on the interpretation of this initialization.  
`rcScopedDelete<rcContourHole> holes = (rcContourHole*)rcAlloc(...);`
is being interpreted as a copy assignment operator invocation for `rcScopedDelete` in gcc, while MSVC treates it as a call to the `rcScopedDelete(T*)` constructor.

My instinct is to say that MSVC is wrong here.  Options are to change the constructor call to
`rcScopedDelete<rcContourHole> holes((rcContourHole*)rcAlloc(...))`
or add an assignment operator with the signature `rcScopedDelete<T>& rcScopedDelete<T>::operator=(T*)`
or define the copy assignment operator and constructor for rcScopedDelete.  Thoughts?

grahamboree:
Alright, so according to http://stackoverflow.com/a/1051468 MSVC is mis-interpreting the overload resolution rules here, and being more lenient than it should be.  I think the fix for this is either removing the disabled copy constructor and copy assignment operator (and hope noone tries to copy an `rcScopedDelete` :worried:) or change the copy-initialization to direct-initialization.  I'm much more in favor of the second, so I'll go ahead and add it to this PR.  FWIW If we were using C++11, we could fix this with a move constructor in the same way `std::unique_ptr` does, since the raw pointer is converted into an rvalue instance of `rcScopedDelete`.

hymerman:
LGTM :+1: Could you squash the commits?

Direct initialisation is the clear winner for me!

Shame we can't use C++11's `=delete` :(

rafzi:
Why did you not remove the empty virtual destructors in subclasses? They do not need to be explicitly defined as well.

grahamboree:
@rafzi Implicitly-defined destructors are never virtual, and even though it's inconsequential in these specific instances, I think it's easier to read and reason about the code with an empty virtual destructor.  i.e. "Okay this destructor doesn't do anything, but it's a subclass" vs "Wait, there's no destructor and it's a subclass.  Is that a mistake?"

grahamboree:
@rafzi Actually it looks like this was changed (for the better) in C++11 such that implicitly-defined destructors are virtual if the base class's destructor is also virtual.  Nice!  Where R&D might not be compiled with C++11, though, I'd rather explicitly define a virtual empty destructor just to be safe.

hymerman:
Implicitly defined destructors are virtual if the base class' destructor is virtual, even before C++11. I just did this quick test to demonstrate: http://cpp.sh/2p4d

@rafzi is right, they can be removed :)

grahamboree:
@hymerman @rafzi TIL :innocent:

hymerman:
Looks even better now :)

Issue #167, opened 2016-01-20
watershed partitioning region id overflow goes undetected


wbierman:
rcBuildRegions() is capable of overflowing regionId.  This goes undetected and can cause a very hard to identify crash later.  I realize that to fix this properly may require a larger rework than you want to do, but if you do not want to widen the region id variable(s), I would suggest you at least add a hard failure in this case so people do not have to chase this bug again.

I can submit a PR which will detect this scenario, fail, and log it, if you would like.

Edit: Just to be clear, I myself would prefer that these variables be widened to unsigned int, rather than unsigned short. :+1:

hymerman:
Yes please Bill - that sounds very useful!

wbierman:
Which way do you want to go?  Log and fail?  Or widen region IDs?

hymerman:
Log and fail - I'm a big fan of making errors more visible! Widening region
IDs is something we could make optional later (perhaps configured at
compile-time in some kind of config.h file).

wbierman:
Yes.  I was assuming I would use a typedef for the new type.  But okay, log and fail it is.  I was able to work around this occurrence of the problem in my own case by increasing cell size a bit.

PR #169, opened 2016-01-21
Detect, log and fail in the event of an overflow of region ids in watershed partitioning

wbierman:


grahamboree:
This is a fix discussed in #167

grahamboree:
LGTM.  Thanks!  :+1:

PR #179, opened 2016-02-15
Fail when too many nodes are requested

### LINKED ISSUE #178: dtNavMeshQuery::init should fail on too many nodes requested
Otherwise path finds may result in invalid paths or infinite loops.
This should be checked as discussed in #64.

Janiels:
dtNavMeshQuery now fails initialization if too many nodes are requested
in the node pool. This could cause wrong paths and infinite loops to
happen if the node indices started overflowing dtNodeIndex or
dtNode::pidx.

Fix #178

### DIFF
--- a/Detour/Include/DetourNavMeshQuery.h
+++ b/Detour/Include/DetourNavMeshQuery.h
@@ -172,7 +172,7 @@ public:
 	
 	/// Initializes the query object.
 	///  @param[in]		nav			Pointer to the dtNavMesh object to use for all queries.
-	///  @param[in]		maxNodes	Maximum number of search nodes. [Limits: 0 < value <= 65536]
+	///  @param[in]		maxNodes	Maximum number of search nodes. [Limits: 0 < value <= 65535]
 	/// @returns The status flags for the query.
 	dtStatus init(const dtNavMesh* nav, const int maxNodes);
 	
--- a/Detour/Include/DetourNode.h
+++ b/Detour/Include/DetourNode.h
@@ -31,18 +31,20 @@ enum dtNodeFlags
 typedef unsigned short dtNodeIndex;
 static const dtNodeIndex DT_NULL_IDX = (dtNodeIndex)~0;
 
+static const int DT_NODE_PARENT_BITS = 24;
+static const int DT_NODE_STATE_BITS = 2;
 struct dtNode
 {
-	float pos[3];				///< Position of the node.
-	float cost;					///< Cost from previous node to current node.
-	float total;				///< Cost up to the node.
-	unsigned int pidx : 24;		///< Index to parent node.
-	unsigned int state : 2;		///< extra state information. A polyRef can have multiple nodes with different extra info. see DT_MAX_STATES_PER_NODE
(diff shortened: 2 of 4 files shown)


hymerman:
LGTM.

I guess this constitutes a breaking change since the range of `maxNodes` has been reduced by 1, is that right? Or did it not work at 65536 before either?

Janiels:
65536 did not work before either (when `dtNodeIndex` was `unsigned short`) as the node with index `65535 = DT_NULL_IDX` cannot correctly be used.

But yes it's actually a pretty significant breaking change; if you were passing a value of >= 0x10000 as `maxNodes` all queries would work as long as they used `< 0x10000` nodes internally (and this is probably the vast majority of queries happening on any nav mesh). Once a query used more than those nodes it would silently give incorrect results or go into infinite loops (as noted in #64).

If you guys think this is a big problem we can hold off with the change until the next version; however, in any case, passing `maxNodes >= 0x10000` is a problem which is silently ignored until this is merged, even if it works in the common cases.

(Bear in mind that my post assumes `dtNodeIndex` is `unsigned short`; if it has been widenened then there were only possible problems if `maxNodes >= (1 << 24) - 1`).

grahamboree:
IMO the breaking change here is acceptable.  It sounds like if you're making requests with maxNodes >= 0x10000, you're running a minor risk of very bad things happening.  As a user, I think I'd rather have the request outright fail immediately than have the potential (however small) of returning bad data or (even worse) infinite looping.  I'd rather catch the invalid parameter bug immediately than try to triage it on my own if/when it goes wrong.

hymerman:
It's definitely acceptable - I'm just wondering if this is for 1.5.1, 1.6.0 or 2.0.0 :)

Seems to me that since it still works for the same range of documented values, this shouldn't be considered breaking - the precondition on the parameter is part of the public API. The only thing that could be considered breaking is that someone might have picked exactly 65536 for `maxNodes` since that was the documented maximum (and the fact it didn't work was a bug)... but I suppose we could call that a documentation bug and that the API really should have disallowed that?

Maybe I'm over-thinking it!

Janiels:
In the sense that it was documented like that it's not as breaking - but in the sense that it worked fine in the majority of cases before when `maxNodes >= 0x10000` is a little more breaking.

So if we followed semantic versioning to the dot it would probably only go for 2.0.0. But it can also be argued that we didn't have "compatibility" with `maxNodes >= 0x10000` already; so in that sense there is still full backwards compatibility. I suppose we can put it in 1.6.0?

grahamboree:
I like 1.6 and maybe even 1.5.1.  Seems like the dominating aspect of this is that it's a low-frequency crash bug fix, which IMO feels critical enough to get it out there sooner than later.

hymerman:
I've done something a little different... I rebased your changes onto master (just because history is starting to look hairy with all these merge commits) and cherry-picked them into a new 1.5.x patch release branch. GitHub won't recognise the rebased commit as it has a different hash so I'll close this manually - it's still in the repo though :)

Janiels:
Sounds good. :+1:

PR #180, opened 2016-02-15
Fix warnings and increase demo text pool size

Janiels:
Fix three warnings added in fc5df2c, and increase the text pool size in
the demo. This would frequently be overflowed, even just by building
with the default settings and expanding the log.

### DIFF
--- a/Detour/Source/DetourNavMeshQuery.cpp
+++ b/Detour/Source/DetourNavMeshQuery.cpp
@@ -716,6 +716,8 @@ public:
 
 	void process(const dtMeshTile* tile, dtPoly** polys, dtPolyRef* refs, int count)
 	{
+		dtIgnoreUnused(polys);
+
 		for (int i = 0; i < count; ++i)
 		{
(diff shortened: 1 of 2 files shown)


Detour/Source/DetourNavMeshQuery.cpp
```diff
@@ -714,7 +714,7 @@ class dtFindNearestPolyQuery : public dtPolyQuery
 	dtPolyRef nearestRef() const { return m_nearestRef; }
 	const float* nearestPoint() const { return m_nearestPoint; }
 
-	void process(const dtMeshTile* tile, dtPoly** polys, dtPolyRef* refs, int count)
+	void process(const dtMeshTile* tile, dtPoly** /*polys*/, dtPolyRef* refs, int count)
```
grahamboree: What about dtIgnoreUnused?

Janiels: Good idea! Didn't know about that one.

hymerman:
:ship:

PR #181, opened 2016-02-16
Update stb_image and stb_truetype to latest versions to fix (most) warnings

hymerman:
Just thought this would be a good idea as they've both been improved over the years, to the point where nearly all the warnings are gone. Tested RecastDemo on Windows and it works fine.


Janiels:
LGTM :+1:

grahamboree:
So, I'm all for updating this to the latest, but do we really need the whole Slideshow system in RecastDemo at all?  Seems like it's leftover from some of Mikko's presentations?  Might be able to just remove this stuff altogether.

hymerman:
Good point - we should get rid of that. stb_truetype is still used elsewhere though - stb_image is only used by `SlideShow` so it could be removed along with it.

hymerman:
Removing `SlideShow` in #184, and upgrading just stb_truetype in #183. I'll close this issue now.

grahamboree:
:+1:

PR #186, opened 2016-02-22
Fix Win64 when populating input file list

### LINKED ISSUE #185: Crash in RecastDemo, x64, VS2015
Filelist.cpp
Function: scanDirectoryAppend
Line: 38
`long fh = _findfirst(pathWithExt.c_str(), &dir);`
According to [MSDN](https://msdn.microsoft.com/ru-ru/library/zyzxfzac.aspx) fh should be 'intptr_t':
`intptr_t fh = _findfirst(pathWithExt.c_str(), &dir);`
With this fix i'm able to "Choose mesh" without a crash. Sorry for my bad English ;-)

hymerman:
As identified by @b-desconocido, `_findfirst` has a return type of `intptr_t`;
this was previously being truncated to long, resulting in a crash on 64 bit Windows.

Fix #185

### DIFF
--- a/RecastDemo/Source/Filelist.cpp
+++ b/RecastDemo/Source/Filelist.cpp
@@ -35,7 +35,7 @@ void scanDirectoryAppend(const string& path, const string& ext, vector<string>& 
 	string pathWithExt = path + "/*" + ext;
 	
 	_finddata_t dir;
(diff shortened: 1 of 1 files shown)


grahamboree:
LGTM :+1:

PR #187, opened 2016-02-22
Fix crashes and odd behaviour in RecastDemo tests

hymerman:
Would crash if selecting a test with geometry then a test with missing geometry
Would use the previous sample type if selecting a sample that doesn't exist in the list
Two of the test cases had an incorrect sample name


RecastDemo/Source/main.cpp
```diff
@@ -817,33 +819,30 @@ int main(int /*argc*/, char** /*argv*/)
...
+						cameraPos[0] = (bmax[0] + bmin[0]) / 2 + camr;
+						cameraPos[1] = (bmax[1] + bmin[1]) / 2 + camr;
+						cameraPos[2] = (bmax[2] + bmin[2]) / 2 + camr;
+						camr *= 3;
+					}
+					cameraEulers[0] = 45;
+					cameraEulers[1] = -45;
+					glFogf(GL_FOG_START, camr * 0.2f);
+					glFogf(GL_FOG_END, camr * 1.25f);
+
```
grahamboree: This block is copy/pasted in 2 other places.  Should those be updated as well?  Lines 636 and 708

hymerman: Well spotted! They would also break if the selected geometry failed to load, although they shouldn't fail to create the sample as that's not data-driven as it is in the tests. I'll make the change.

hymerman: Turns out there was only one other change needed; if after choosing geometry, that geometry failed to load, the previous geometry would get deleted and the sample would end up with a dangling pointer.

grahamboree:
Cool :+1:

PR #188, opened 2016-02-26
Unit tests for a good number of core Recast functions.

grahamboree:
These are by no means complete coverage of the functions they test, but are at least a start.

### DIFF
--- a/Tests/Recast/Tests_Recast.cpp
+++ b/Tests/Recast/Tests_Recast.cpp
@@ -2,6 +2,117 @@
 
 #include "Recast.h"
 
+TEST_CASE("rcSwap")
+{
+	SECTION("Swap two values")
+	{
+		int one = 1;
+		int two = 2;
+		rcSwap(one, two);
+		REQUIRE(one == 2);
+		REQUIRE(two == 1);
+	}
+}
+
+TEST_CASE("rcMin")
+{
+	SECTION("Min returns the lowest value.")
+	{
+		REQUIRE(rcMin(1, 2) == 1);
+		REQUIRE(rcMin(2, 1) == 1);
+	}
+
+	SECTION("Min with equal args")
+	{
+		REQUIRE(rcMin(1, 1) == 1);
+	}
+}
+
+TEST_CASE("rcMax")
+{
+	SECTION("Max returns the greatest value.")
+	{
+		REQUIRE(rcMax(1, 2) == 2);
+		REQUIRE(rcMax(2, 1) == 2);
+	}
+
+	SECTION("Max with equal args")
+	{
+		REQUIRE(rcMax(1, 1) == 1);
+	}
+}
+
+TEST_CASE("rcAbs")
+{
+	SECTION("Abs returns the absolute value.")
+	{
+		REQUIRE(rcAbs(-1) == 1);
+		REQUIRE(rcAbs(1) == 1);
+		REQUIRE(rcAbs(0) == 0);
+	}
+}
+
+TEST_CASE("rcSqr")
+{
+	SECTION("Sqr squares a number")
+	{
+		REQUIRE(rcSqr(2) == 4);
+		REQUIRE(rcSqr(-4) == 16);
+		REQUIRE(rcSqr(0) == 0);
+	}
+}
+
+TEST_CASE("rcClamp")
+{
+	SECTION("Higher than range")
+	{
+		REQUIRE(rcClamp(2, 0, 1) == 1);
+	}
+
+	SECTION("Within range")
+	{
+		REQUIRE(rcClamp(1, 0, 2) == 1);
+	}
+
+	SECTION("Lower than range")
+	{
+		REQUIRE(rcClamp(0, 1, 2) == 1);
+	}
+}
+
+TEST_CASE("rcSqrt")
+{
+	SECTION("Sqrt gets the sqrt of a number")
+	{
+		REQUIRE(rcSqrt(4) == Approx(2));
+		REQUIRE(rcSqrt(81) == Approx(9));
+	}
+}
+
+TEST_CASE("rcVcross")
+{
+	SECTION("Computes cross product")
+	{
+		float v1[3] = {3, -3, 1};
+		float v2[3] = {4, 9, 2};
+		float result[3];
+		rcVcross(result, v1, v2);
+		REQUIRE(result[0] == Approx(-15));
+		REQUIRE(result[1] == Approx(-2));
+		REQUIRE(result[2] == Approx(39));
+	}
+
+	SECTION("Cross product with itself is zero")
+	{
+		float v1[3] = {3, -3, 1};
+		float result[3];
+		rcVcross(result, v1, v1);
+		REQUIRE(result[0] == Approx(0));
+		REQUIRE(result[1] == Approx(0));
+		REQUIRE(result[2] == Approx(0));
+	}
+}
+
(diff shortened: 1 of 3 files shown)


Tests/Recast/Tests_Recast.cpp
```diff
@@ -1,6 +1,118 @@
...
+{
+	SECTION("Max returns the greatest value.")
+	{
+		REQUIRE(rcMax(1, 2) == 2);
+		REQUIRE(rcMax(2, 1) == 2);
+	}
+
+	SECTION("Max with equal args")
+	{
+		REQUIRE(rcMin(1, 1) == 1);
```
Janiels: Looks like a copy paste mistake here.

Tests/Recast/Tests_Recast.cpp
```diff
@@ -1,6 +1,118 @@
...
+		REQUIRE(rcClamp(0, 1, 2) == 1);
+	}
+}
+
+TEST_CASE("rcSqrt")
+{
+	SECTION("Sqrt gets the sqrt of a number")
+	{
+		REQUIRE(rcSqrt(4) == 2);
+		REQUIRE(rcSqrt(81) == 9);
```
Janiels: Should `Approx` be used here? (like below)

Tests/Recast/Tests_Recast.cpp
```diff
@@ -10,5 +122,715 @@ TEST_CASE("rcVdot")
...
+		float v1[3] = {1, 2, 3};
+		float v2[3] = {0, 2, 4};
+		float result[3];
+		rcVmad(result, v1, v2, 2);
+		REQUIRE(result[0] == Approx(1));
+		REQUIRE(result[1] == Approx(6));
+		REQUIRE(result[2] == Approx(11));
+	}
+
+	SECTION("first vector is scaled, second is not")
```
Janiels: Should this say "second vector is scaled, first is not"?

Tests/Recast/Tests_Recast.cpp
```diff
@@ -10,5 +122,715 @@ TEST_CASE("rcVdot")
...
+	{
+		float v1[3] = {5, 4, 0};
+		float v2[3] = {1, 2, 9};
+		rcVmax(v1, v2);
+		REQUIRE(v1[0] == 5);
+		REQUIRE(v1[1] == 4);
+		REQUIRE(v1[2] == 9);
+	}
+
+	SECTION("v1 is max")
```
Janiels: Should be "v2 is max"

Tests/Recast/Tests_Recast.cpp
```diff
@@ -10,5 +122,715 @@ TEST_CASE("rcVdot")
...
+	{
+		float v1[3] = {1, 2, 3};
+		float v2[3] = {4, 5, 6};
+		rcVmax(v1, v2);
+		REQUIRE(v1[0] == 4);
+		REQUIRE(v1[1] == 5);
+		REQUIRE(v1[2] == 6);
+	}
+
+	SECTION("v2 is max")
```
Janiels: Should be "v1 is max"

Tests/Recast/Tests_Recast.cpp
```diff
@@ -10,5 +122,715 @@ TEST_CASE("rcVdot")
...
+
+	SECTION("Add a span to an empty heightfield.")
+	{
+		bool result = rcAddSpan(&ctx, hf, x, y, smin, smax, area, flagMergeThr);
+		REQUIRE(result);
+		REQUIRE(hf.spans[0] != 0);
+		REQUIRE(hf.spans[0]->smin == smin);
+		REQUIRE(hf.spans[0]->smax == smax);
+		REQUIRE(hf.spans[0]->area == area);
+		REQUIRE(hf.spans[0]->area == area);
```
Janiels: Duplicate line

Tests/Recast/Tests_Recast.cpp
```diff
@@ -10,5 +122,715 @@ TEST_CASE("rcVdot")
...
+
+	SECTION("Add a span that gets merged with an existing span.")
+	{
+		bool result = rcAddSpan(&ctx, hf, x, y, smin, smax, area, flagMergeThr);
+		REQUIRE(result);
+		REQUIRE(hf.spans[0] != 0);
+		REQUIRE(hf.spans[0]->smin == smin);
+		REQUIRE(hf.spans[0]->smax == smax);
+		REQUIRE(hf.spans[0]->area == area);
+		REQUIRE(hf.spans[0]->area == area);
```
Janiels: Duplicate line

Janiels:
I left a few comments. There are many places where `Approx` is used, and many places where it isn't - should it be used consistently for all the floating point comparisons?
Apart from that the rest LGTM.

Tests/Recast/Tests_Recast.cpp
grahamboree: ah! good catch :+1:

grahamboree:
Thanks again for the code review @Janiels.  I fixed the things you pointed out and made all floating point comparisons use Approx for consistency, even if they don't need it.  I think it's clearer in the long run, and expresses the intent more clearly.  :smile:

Janiels:
I notice that the created heightfields are leaked after the tests are over. Is this something we are OK with to promote the clarity of the code?
It might be worth it to use a linear allocator that automatically cleans up after each test case through RAII. Otherwise, as the test suite grows, we might start using a lot of memory.

hymerman:
Good point @Janiels - `rcHeightfield` has both `rcAllocHeightfield` and `rcCreateHeightfield` which roughly correspond to `new` and a constructor. There's `rcFreeHeightfield` to pair with `rcAllocHeightfield`, which roughly corresponds with `delete`, but no such thing as `rcDestroyHeightfield` which would be like the destructor, and which would free `spans`. So `rcHeightfield` can't be used on the stack at the moment without leaking memory (or manually freeing its internals!). That's definitely something we should fix! But for now, I'd suggest allocating the heightfields on the heap and freeing them at the end of the tests - does Catch allow test tear-down or will this have to be repeated in each `SECTION`? Or even better, if we're ok with using C++11 in tests (I am), use `unique_ptr` with a custom deleter.

A couple of things for the future (we can tweak them later, no sense holding up the PR for them) would be use of `const` for test-local things, and again if we're ok with C++11, `nullptr` instead of `0`. They're super minor things though.

Apart from that, I'm happy with these tests - a great start! Thanks!

grahamboree:
I messed around with std::unique_ptr, which I was able to get to automatically call rcFreeHeightfield, however, this required getting travis and appveyor to build C++11 for the RecastDemo project, which proved to be a bit of a hassle.  I still think it's useful to get them using C++11, however, I also think that allowing for stack-allocated rcHeightfields is more useful.  
  Especially where users are expected to directly interact with rcHeightfields, it's a pretty nasty pitfall that stack-allocated heightfields will leak memory.  I think a better approach is to allow for stack-allocations as well.  I submitted this change separately as #196, so we can discuss the implications of this change there, but this PR will have to wait until either #196, or I get C++11 build configs working well enough to merge in.

hymerman:
Sounds good - thanks Graham!

grahamboree:
Should be g2g with the new ctor/dtor on rcHeightfield, barring any further suggestions ;)

Janiels:
LGTM

hymerman:
:+1:

PR #192, opened 2016-03-13
Miscellaneous warning fixes, again

hymerman:
Yet more warning fixes for a few things - dead assignments throughout the codebase as reported in #93.

Fixes #93.

### DIFF
--- a/Detour/Include/DetourCommon.h
+++ b/Detour/Include/DetourCommon.h
@@ -20,6 +20,7 @@
 #define DETOURCOMMON_H
 
 #include "DetourMath.h"
+#include <stddef.h>
 
 /**
 @defgroup detour Detour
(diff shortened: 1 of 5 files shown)


Detour/Include/DetourCommon.h
```diff
@@ -482,6 +483,15 @@ inline void dtSwapEndian(float* v)
 void dtRandomPointInConvexPoly(const float* pts, const int npts, float* areas,
 							   const float s, const float t, float* out);
 
+template<typename TypeToRetrieveAs, typename BufferType>
+TypeToRetrieveAs* getThenAdvanceBufferPointer(BufferType*& buffer, const size_t distanceToAdvance)
```
Janiels: Other global Detour functions are prefixed by `dt`. I think we should do the same here.
Does `buffer` need to be a templated type?

Janiels:
I left some feedback. Apart from that, looks good.

If we switch to C++11 we might also want to statically assert std::is_trivially_copyable in `getThenAdvanceBufferPointer`. I believe we do need C++11 or some tricks for that though.

Detour/Include/DetourCommon.h
hymerman: Good point - I'll prefix with `dt`.

`BufferType` doesn't strictly need to be a template argument but if I leave the parameter as `void*&` then every call site needs to cast to `void*`, which is quite cumbersome - the template just helps with readability.

hymerman:
Thanks for the feedback - I'll update in a jiffy!

No need to assert for that here, as it's just advancing a pointer - the things in the buffer can be whatever they like as they're not getting modified.

Detour/Include/DetourCommon.h
Janiels: Couldn't it be `unsigned char*&`? `buffer` is always passed as `unsigned char *` as far as I can see.

Janiels:
I meant that `TypeToRetrieveAs` must have `std::is_trivially_copyable`, otherwise reinterpreting the buffer as that type is undefined behavior, I believe.

hymerman:
Ah, I'm not aware of that rule - I thought it would only be undefined on writing to (or copying out of) the buffer - not on access. But I suppose if you're accessing it using `dtGetThenAdvanceBufferPointer` then someone must have already copied it in, so an assert would be helpful either way!

Something I'd like to start doing eventually is using some C++11/14 features behind macro guards to allow pre-C++11 compilers to still build the code - imagine a "`RC_OVERRIDE`" and "`RC_CONSTEXPR`". `static_assert`ing in places like this would be another great addition.

Detour/Include/DetourCommon.h
hymerman: It could, yes - I thought I saw places using `signed char*` which would also be valid but after a second look you're right, it's only ever `unsigned char*`. I'll remove the template argument for simplicity.

hymerman: Quick note: I removed the template argument but had to keep two overloads to support `const` buffers.

PR #193, opened 2016-03-14
Miscellaneous warning fixes, part 3

hymerman:
Yet more warning fixes, for a smattering of issues found using either Coverity Scan, clang's scan-build or ReSharper C++. Each fix is in its own commit so the best way to see what's fixed is to read the individual commit messages.

### DIFF
--- a/RecastDemo/Source/SampleInterfaces.cpp
+++ b/RecastDemo/Source/SampleInterfaces.cpp
@@ -20,6 +20,8 @@ BuildContext::BuildContext() :
 	m_messageCount(0),
 	m_textPoolSize(0)
 {
+	memset(m_messages, 0, sizeof(char*) * MAX_MESSAGES);
+
 	resetTimers();
 }
 
--- a/RecastDemo/Source/TestCase.cpp
+++ b/RecastDemo/Source/TestCase.cpp
@@ -102,9 +102,22 @@ bool TestCase::load(const std::string& filePath)
 	FILE* fp = fopen(filePath.c_str(), "rb");
 	if (!fp)
 		return false;
-	fseek(fp, 0, SEEK_END);
-	int bufSize = ftell(fp);
-	fseek(fp, 0, SEEK_SET);
+	if (fseek(fp, 0, SEEK_END) != 0)
+	{
+		fclose(fp);
+		return false;
+	}
+	long bufSize = ftell(fp);
+	if (bufSize < 0)
+	{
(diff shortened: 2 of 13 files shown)


RecastDemo/Source/SampleInterfaces.cpp
```diff
@@ -20,6 +20,8 @@ BuildContext::BuildContext() :
 	m_messageCount(0),
 	m_textPoolSize(0)
 {
+	memset(m_messages, 0, sizeof(char) * MAX_MESSAGES);
```
Janiels: Should be `sizeof(char*) * MAX_MESSAGES`

RecastDemo/Source/TestCase.cpp
```diff
@@ -102,9 +102,22 @@ bool TestCase::load(const std::string& filePath)
 	FILE* fp = fopen(filePath.c_str(), "rb");
 	if (!fp)
 		return false;
-	fseek(fp, 0, SEEK_END);
+	if (fseek(fp, 0, SEEK_END) != 0)
+	{
+		fclose(fp);
+		return false;
+	}
 	int bufSize = ftell(fp);
```
Janiels: I think we should change this to be of type `long` now that you're making changes around here. On 64-bit Linux distros I seem to recall that `long` is frequently 8 bytes and `int` is 4, and `ftell` returns `long`.

Janiels:
Left some feedback, otherwise LGTM. Maybe add asserts to the places we now rely on `nv > 0`, but that's pretty minor and should never be hit anyway.

RecastDemo/Source/SampleInterfaces.cpp
hymerman: Well spotted! We should definitely get 64 bit builds up and running to catch this kind of thing.

RecastDemo/Source/TestCase.cpp
hymerman: True - I'll update them. And again, a warning-free 64 bit build will help with this stuff - I'll have a look at doing that soon.

hymerman:
Thanks for the feedback Jakob! I've made the changes.

I wasn't going to squash since these are all separate independent issues, just lumped into one PR to make them easier to manage. I have rebased the second round of changes into existing commits though.

Janiels:
I see, that makes sense, especially with the commit messages being as descriptive as they are.
LGTM.

PR #194, opened 2016-03-15
Fail when too many nodes are requested

### LINKED ISSUE #178: dtNavMeshQuery::init should fail on too many nodes requested

coovig:
dtNavMeshQuery now fails initialization if too many nodes are requested
in the node pool. This could cause wrong paths and infinite loops to
happen if the node indices started overflowing dtNodeIndex or
dtNode::pidx.

Fix #178

### DIFF
--- a/Detour/Include/DetourNavMeshQuery.h
+++ b/Detour/Include/DetourNavMeshQuery.h
@@ -161,7 +161,7 @@ public:
 	
 	/// Initializes the query object.
 	///  @param[in]		nav			Pointer to the dtNavMesh object to use for all queries.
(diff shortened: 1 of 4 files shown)


hymerman:
I'm not sure what this PR is for; this change is already present in the master branch (it was cherry-picked so has a different hash). I don't know why GitHub is allowing this.

See here:
https://github.com/recastnavigation/recastnavigation/commit/09afa02f4a51c4d55df58b8640b46026d195af8d

I'll close this issue now, but please do reply (mentioning me) if you think this is wrong.

PR #196, opened 2016-03-20
Added Constructor and Destructor to rcHeightfield

grahamboree:
This adds a default constructor and a destructor to the rcHeightfield struct, and disables the default copy constructor and copy assignment operator.  

While implementing unit tests that use rcHeightfield, there was an observation that rcHeightfield's that aren't made through `rcAllocHeightfield()` and subsequently free'd thorugh `rcFreeHeightfield()` leak memory.  This restricted all rcHeightfield allocations to the heap.  In addition to being an unnecessary constraint that could have adverse performance impacts, it also was causing memory leaks in test cases that couldn't call `rcFreeHeightfield()`.

This shouldn't have any measurable impact on performance, as heightfield's aren't allocated nearly as often as other objects, and the compiler should do a good job of inlining calls to the constructor and destructor where appropriate.

### DIFF
--- a/Recast/Include/Recast.h
+++ b/Recast/Include/Recast.h
@@ -293,6 +293,9 @@ struct rcSpanPool
 /// @ingroup recast
 struct rcHeightfield
 {
+	rcHeightfield();
+	~rcHeightfield();
+
 	int width;			///< The width of the heightfield. (Along the x-axis in cell units.)
 	int height;			///< The height of the heightfield. (Along the z-axis in cell units.)
 	float bmin[3];  	///< The minimum bounds in world space. [(x, y, z)]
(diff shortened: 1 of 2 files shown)


Janiels:
LGTM

hymerman:
I don't have the codebase cloned here so can't check right now, but are there any spots where `memset` is called on an `rcHeightfield`? If so those should be changed too since `rcHeightfield` is no longer Trivially Copyable, which makes `memset`ting them undefined behaviour.

Otherwise though, this is good to go!

grahamboree:
Just looked through the codebase and couldn't find any other instances where an rcHeightfield was being memset-ed.  Is this something coverity would catch?

hymerman:
Coverity or some other static analysis tool probably would catch it, yes. If you can't find anything then let's go ahead and merge this :) :+1:

PR #200, opened 2016-04-03
Fix memory leak in LinearAllocator used by Sample_TempObstacles

Sandern:
Caused by #158.

Can be reproduced by selecting the "Temp Obstacles" sample and pressing build multiple times. Memory will leak because it uses the alloc function from dtTileCacheAlloc and the free function from LinearAllocator (which is empty).

Another improvement would be to change dtTileCacheAlloc to an abstract empty implementation, similar to dtTileCacheCompressor. Existing usage of dtTileCacheAlloc would break though.


Janiels:
Thanks!

hymerman:
Adding `DT_OVERRIDE` has just jumped up my priority list a little...

PR #202, opened 2016-04-19
Fixed a links issue due to offmesh connections

tfabretti:
When creating the tile's links, offmesh connections inside the tile
were added twice, leading to the array of links being full and not
being able to store the links computed after (portals between
tiles).

Google group discussion : 
https://groups.google.com/forum/#!msg/recastnavigation/l-_Mk1O4I04/PWE0dGSKBAAJ

### DIFF
--- a/Detour/Source/DetourNavMesh.cpp
+++ b/Detour/Source/DetourNavMesh.cpp
@@ -943,7 +943,10 @@ dtStatus dtNavMesh::addTile(unsigned char* data, int dataSize, int flags,
 	tile->flags = flags;
 
 	connectIntLinks(tile);
+
+	// Base off-mesh connections to their starting polygons and connect connections inside the tile.
 	baseOffMeshLinks(tile);
+	connectExtOffMeshLinks(tile, tile, -1);
 
 	// Create connections with neighbour tiles.
 	static const int MAX_NEIS = 32;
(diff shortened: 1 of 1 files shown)


Detour/Source/DetourNavMesh.cpp
```diff
@@ -958,9 +958,9 @@ dtStatus dtNavMesh::addTile(unsigned char* data, int dataSize, int flags,
 		{
 			connectExtLinks(tile, neis[j], -1);
 			connectExtLinks(neis[j], tile, -1);
+			connectExtOffMeshLinks(neis[j], tile, -1);
 		}
 		connectExtOffMeshLinks(tile, neis[j], -1);
-		connectExtOffMeshLinks(neis[j], tile, -1);
```
memononen: Maybe it would be more clear to explicitly connect the links within a tile, and then specifically per layer. That is:
```C++
	connectIntLinks(tile);

	// Base off-mesh connections to their starting polygons and connect connections inside the tile.
	baseOffMeshLinks(tile);
	connectExtOffMeshLinks(tile, tile, -1);

	// Create connections with neighbour tiles.
	static const int MAX_NEIS = 32;
	dtMeshTile* neis[MAX_NEIS];
	int nneis;
	
	// Connect with layers in current tile.
	nneis = getTilesAt(header->x, header->y, neis, MAX_NEIS);
	{
		if (neis[j] == tile)
			continue;
		connectExtLinks(tile, neis[j], -1);
		connectExtLinks(neis[j], tile, -1);
		connectExtOffMeshLinks(tile, neis[j], -1);
		connectExtOffMeshLinks(neis[j], tile, -1);
	}
```

I have not tested that, though.

Janiels: I like that, that makes it a lot clearer that the tile<->tile is a special case. @tfabretti could you make and test that change on your mesh?

tfabretti:
I agree, it would be better. I will test your solution tomorrow (since it's just moving a line, it should work but I'll check) and make the change here.

Thanks !

Detour/Source/DetourNavMesh.cpp
tfabretti: I agree ! I'll try it out tomorrow (it's just moving a line and it should work but I'll test it !).

I'll make the change right after the test.

Janiels:
Thanks for the fix! I didn't get a notification about the update, so didn't see it until now.

PR #203, opened 2016-04-25
Add upToDate param for dtTileCache::update

Janiels:
This changes dtTileCache::update to add an optional parameter that
indicates whether the tile cache is up to date. This is useful to
determine whether calling the function again would do anything.


hymerman:
Looks good - maybe just add a comment explaining the `upToDate` parameter?

Janiels:
Good idea, added!

hymerman:
Nice; :+1: from me! I'll leave it to you to merge how you like, or wait for more input :)

PR #205, opened 2016-05-18
Append results correctly in Dijkstra searches

### LINKED ISSUE #204: findPolysAroundCircle and findPolysAroundShape do not update results when paths improve
The implementations of these two Dijkstra searches add polygons to the results when they are seen as neighbors to the current best node. This is problematic because a shorter path can potentially be found later during the search, in which case the path will be updated on the open list.

Instead we should append these polygons to the results once we process them. Dijkstra's algorithm guarantees that when we process a node, we will never process it again. In other words we have found the shortest path to a node once we move it into the closed list, so at this point it can be appended.

Janiels:
This changes the Dijkstra searches in findPolysAroundCircle and
findPolysAroundShape to append the results when we visit nodes in the
main loop. The algorithm guarantees that we will never visit them again
after this, so at this point we are guaranteed that we cannot find a
better path.

Previously results would be appended the first time we saw a polygon
which meant later improvements on the paths to those polygons would not
be reflected in the results.

Fix #204

### DIFF
--- a/Detour/Source/DetourNavMeshQuery.cpp
+++ b/Detour/Source/DetourNavMeshQuery.cpp
@@ -2743,20 +2743,6 @@ dtStatus dtNavMeshQuery::findPolysAroundCircle(dtPolyRef startRef, const float* 
 	dtStatus status = DT_SUCCESS;
 	
 	int n = 0;
-	if (n < maxResult)
-	{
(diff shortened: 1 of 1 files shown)


grahamboree:
LGTM :+1:

PR #207, opened 2016-05-18
Add parent indices in Dijkstra searches

### LINKED ISSUE #206: Improve usability of findPolysAroundCircle and findPolysAroundShape by returning parent indices
Right now these methods return as output the parent poly refs of each polygon. As an example, `resultRef[1]` is some polygon, and `resultParent[1]` is the polygon before `resultRef[1]` on the shortest path to `resultRef[1]`.
However, since these are returned as poly refs it makes it very inconvenient to retrace paths, having to resort to something like this:
```cpp
dtPolyRef polys[128];
dtPolyRef polyParents[128];

m_query->findPolysAroundCircle(..., polys, polyParents, ...);

std::vector<dtPolyRef> path;
dtPolyRef poly = ...; // some poly we want to find the shortest path to
while (poly)
{
  path.insert(path.begin(), poly);
  poly = 0;
  for (int i = 0; i < numPolys; i++)
  {
    if (polys[i] == poly)
    {
      poly = polyParents[i];
      break;
    }
  }
}
```

The problem is finding the index of the parent poly at each step, which requires a linear search when the return values are given this way. Instead the API should output indices into the `resultRef` array for the parent polys, which could alleviate this cost.

Note that it is possible to avoid the linear search but it requires using the node pool Detour used to perform the search.

Janiels:
This changes the `findPolysAroundCircle`, `findPolysAroundShape` and
`findLocalNeighbourhood` functions to return indices into the poly array
for the parents, instead of returning the polygon references. This makes
it much more convenient to retrace paths after performing the searches,
based on the parent arrays.

Also make some arguments to the aforementioned functions required as the
APIs are not very useful without them. Update documentation to specify
this and perform addition parameter validation.

Fix #206

**Note: This change is breaking in the following ways:**
1. The functions now fail if null is passed for `resultRef`. Doing this does not result in anything useful, and having the results simplifies the implementation.
2. `resultParent` from `dtPolyRef` to `int`
3. `dtNode::pidx` now represents indices into the `resultRef` array during and after the Dijkstra searches rather than representing indices into the node pool. This is breaking since the node pool is public.

Of these I think number three is the worst. This implementation has the least overhead, but it's a little iffy since it uses `dtNode::pidx` in a different way than other functions. For that reason I am not actually convinced this change is worth it in the form in this PR since the node pool is publicly exposed and can be used for the scenario I describe in #206.

### DIFF
--- a/Detour/Include/DetourNavMeshQuery.h
+++ b/Detour/Include/DetourNavMeshQuery.h
@@ -267,15 +267,17 @@ public:
 	///  @param[in]		radius			The radius of the search circle.
 	///  @param[in]		filter			The polygon filter to apply to the query.
 	///  @param[out]	resultRef		The reference ids of the polygons touched by the circle. [opt]
-	///  @param[out]	resultParent	The reference ids of the parent polygons for each result. 
-	///  								Zero if a result polygon has no parent. [opt]
+	///  @param[out]	resultParent	The indices into resultRef of the parent polygons for each result. [opt]
+	///  								-1 if a result polygon has no parent (i.e. for resultRef[0]).
+	///  								If resultRef[i] is a polygon, then resultRef[resultParent[i]] is the
+	///  								reference id of the parent polygon.
 	///  @param[out]	resultCost		The search cost from @p centerPos to the polygon. [opt]
 	///  @param[out]	resultCount		The number of polygons found. [opt]
 	///  @param[in]		maxResult		The maximum number of polygons the result arrays can hold.
 	/// @returns The status flags for the query.
 	dtStatus findPolysAroundCircle(dtPolyRef startRef, const float* centerPos, const float radius,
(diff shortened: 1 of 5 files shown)


Janiels:
Closing this out as I don't think the breaking change will be worth it, especially when the scenario described in #206 is doable by using the node pool.

Detour/Include/DetourNavMeshQuery.h
```diff
@@ -267,15 +267,17 @@ class dtNavMeshQuery
 	///  @param[in]		radius			The radius of the search circle.
 	///  @param[in]		filter			The polygon filter to apply to the query.
 	///  @param[out]	resultRef		The reference ids of the polygons touched by the circle. [opt]
```
hymerman: Should take off the '[opt]'

PR #208, opened 2016-05-23
Take area costs into account in Dijkstra searches

Janiels:
`findPolysAroundCircle` and `findPolysAroundShape` were only using distances
to compute the costs. This was inconsistent with other APIs that
performed cost-based searches and accepted filters as parameters.

This is a small breaking change, but the previous behavior is easy to get simply by passing a filter with all area costs as 1. The new behavior is probably what most people would expect from the function once they realize it is a Dijkstra search, so I think we should take this breaking change.


hymerman:
LGTM! The API itself hasn't changed, this feels more like a new feature/improvement ("more queries use filters"), so this feels like it should go in 1.6.

Janiels:
Sounds good to me. Shall I just merge it now or wait until right before 1.6?

hymerman:
Merge it now I think - we've already got a bunch of feature changes since 1.5 so perhaps we should be thinking about a 1.6 soon!

PR #212, opened 2016-06-14
Fix a possible memory leak in dtBuildTileCacheLayer

Janiels:
I've opted not to introduce a RAII class here as `data` is output anyway, so it would mostly benefit for `buffer`; but this file only uses this pattern in this one place, so I don't think it would be worth it.


grahamboree:
Nice catch!  Looks good to me. 👍

hymerman:
Good spot! Weird that Coverity didn't notice this!

jackpoz:
it could be that you need to instruct coverity that dtAlloc() is a custom memory allocator and should always have a dtFree() pair.
You can find more details at https://scan.coverity.com/tune and https://scan.coverity.com/models#c_checker_checkerconfig , in particular
>\__coverity_alloc__() - indicates function allocates memory
\__coverity_free__() - indicates function frees the memory allocated for the pointer

PR #214, opened 2016-07-05
Detect all overflows in rcBuildHeightfieldLayers

Janiels:
The function now fails if it fails to insert all layers because the
array overflowed. Previously this would silently corrupt the heap.

See discussion in #141.

@memononen Can you review this and comment on whether the bail-out on overflow of layers is ok?

### DIFF
--- a/Recast/Source/RecastLayers.cpp
+++ b/Recast/Source/RecastLayers.cpp
@@ -27,7 +27,9 @@
 #include "RecastAssert.h"
 
 
-static const int RC_MAX_LAYERS = RC_NOT_CONNECTED;
+// Must be 255 or smaller (not 256) because layer IDs are stored as
+// a byte where 255 is a special value.
+static const int RC_MAX_LAYERS = 63;
 static const int RC_MAX_NEIS = 16;
 
 struct rcLayerRegion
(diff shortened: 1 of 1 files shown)


hymerman:
LGTM :) But let's wait and see what Mikko says!

memononen:
Code LGTM. While you're at it, I'd recommend to add a better error message, though. Something that allows the user to resolve the problem.

I was thinking to suggesting to increase `RC_NOT_CONNECTED`, but currently it is tied to `RC_NOT_CONNECTED`. I don't think that it is necessary, though, it should be fine to have up to 255 layers (0xff has special meaning).

Another improvement to the error message would be to explain why the error happens (too many overlapping walkable platforms in the input geometry).

Janiels:
@memononen I've made the changes you suggested. Do I need to do anything else to untie it from `RC_NOT_CONNECTED`?

memononen:
@Janiels I recommend to double check in the intermediate format, navmesh builder (i.e. how layer edge connections are encoded), and detour format to see that there are no further limitations to the range. I think they all use at least `char`, but you never know, there might be some special codes and what not :)

Recast/Source/RecastLayers.cpp
```diff
@@ -27,7 +27,8 @@
 #include "RecastAssert.h"
 
 
-static const int RC_MAX_LAYERS = RC_NOT_CONNECTED;
+// Must be 255 or smaller
+static const int RC_MAX_LAYERS = 63;
```
memononen: should leave comment here that max value is 254, as `0xff` is used in some places as "not set" (IIRC).

Janiels: If `RC_MAX_LAYERS` is 255, then the max index will be 254, which should be fine I believe.

Janiels:
@memononen My knowledge of this area of R&D is very limited. As far as I can see the layer IDs do not survive past `rcBuildHeightfieldLayers`, apart from the count of layers themselves. Layer edge connections use 4 bits for internal connections and 4 bits for portal connections - each bit just indicates a side, so that should not rely on the layer IDs. Detour uses the same.

Recast/Source/RecastLayers.cpp
memononen: Correct, my bad :)

memononen:
@Janiels Ok, that sounds correct.

I recommend to remove `RC_NOT_CONNECTED` as you did, because it's just confusing. It could be that there was a reason at some point (maybe I used to store the full index), but I cannot recall it anymore :).

Add a comment what is max value for `RC_MAX_LAYERS`, and we're good to go.

Janiels:
@memononen The comment should already be there, or are you seeing something different?

memononen:
@Janiels got it :) I was expecting it to be on the same line :) If you feel like it, you could add something like: "... because the value is as stored byte, 0xff used in some places as 'no set'."

Anyways, good to go! 👍

Janiels:
Thanks Mikko! Updated the comment and will merge.

PR #215, opened 2016-07-10
Add support for AABB shaped obstacles to dtTileCache

cmf028:
The current dtTileCache supports cylindrical shaped obstacles, but this tends to be a very limiting obstacle shape that doesn't fit many use cases very well. An AABB is a bit more flexible and doesn't take much to implement.

It would be nice to also support extruded convex polys like recast does, but this is a good start.

### DIFF
--- a/DetourTileCache/Source/DetourTileCache.cpp
+++ b/DetourTileCache/Source/DetourTileCache.cpp
@@ -370,9 +370,44 @@ dtStatus dtTileCache::addObstacle(const float* pos, const float radius, const fl
 	memset(ob, 0, sizeof(dtTileCacheObstacle));
 	ob->salt = salt;
 	ob->state = DT_OBSTACLE_PROCESSING;
-	dtVcopy(ob->pos, pos);
-	ob->radius = radius;
-	ob->height = height;
+	ob->type = DT_OBSTACLE_CYLINDER;
+	dtVcopy(ob->cylinder.pos, pos);
+	ob->cylinder.radius = radius;
+	ob->cylinder.height = height;
+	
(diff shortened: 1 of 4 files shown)


hymerman:
Nice and simple. This looks good. :+1:

DetourTileCache/Source/DetourTileCache.cpp
```diff
@@ -370,9 +370,44 @@ dtStatus dtTileCache::addObstacle(const float* pos, const float radius, const fl
...
+	req->action = REQUEST_ADD;
+	req->ref = getObstacleRef(ob);
+	
+	if (result)
+		*result = req->ref;
+	
+	return DT_SUCCESS;
+}
+
+dtStatus dtTileCache::addObstacle(const float* bmin, const float* bmax, dtObstacleRef* result)
```
Janiels: Maybe we can call this something else? I'm not a fan of an overload where the function does something semantically different (well I know, it adds an obstacle, but I mean a box vs a cylinder). Maybe something like addBoxObstacle instead?

Janiels:
I agree, looks good aside from the naming thing I've commented on.

grahamboree:
Agreed.  After the naming change we can merge this.  Thanks! 👍

cmf028:
Changed name as per comments

Janiels:
Thank you, LGTM now. As a last thing, can you fix your Git settings so the commit is properly linked to your Github account?
See here: https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user

grahamboree:
Thanks!

PR #217, opened 2016-07-13
Prune tool start fix

### LINKED ISSUE #216: NavMeshToolPruner::floodNavmesh
start poly isn't being marked
needs
flags->setFlags(start, flag);
at the top of function

Janiels:
The start poly would not be pruned if it was the only poly on the
island.

Fix #216

While I was here I've also taken the liberty of replacing the custom dynamic array with `std::vector`.


hymerman:
:+1: for the fix, and :+1: :+1: for `std::vector` :)

grahamboree:
Another 👍 👍 here for `std::vector`.  😄

PR #218, opened 2016-07-13
Add .editorconfig with tabs

Janiels:
This file specifies the indentation settings used in R&D. This should help keep the code base consistent.
It is very useful when you are switching between multiple projects with different indentations,
or when you have multiple developers with different environment.

Some IDEs support parsing this file format out of the box, and otherwise there exist lightweight plugins that parses the file format and applies the settings for virtually all IDEs.

See http://editorconfig.org/.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+# editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_size = 4
+indent_style = tab


hymerman:
Oh cool, I hadn't heard about this project!

For this kind of thing, in the past I've used clang-format - perhaps we could add a config file for that too? I'd love to run an auto pass over the codebase but it'll make merging existing PRs a bit of a nightmare!

Janiels:
I don't have experience with clang-format, but it certainly sounds useful, and an automatic pass over the code base sounds like a good idea. .editorconfig is not as "beefy". :smile:

I'll go ahead and merge this one then - we can look at adding a clang-format config later.

grahamboree:
Awesome!  I too did not know about this, but it's pretty sweet.  Been wanting something like this for a while. 😄

PR #219, opened 2016-07-13
Add an assertion for alloc in rcIntArray::doResize

### LINKED ISSUE #170: rcIntArray::doResize reallocation failures are ignored
In RecastAlloc.cpp is defined rcIntArray::doResize().  It will reallocate an array twice the previous size and copy contents, much the same way a vector does.

One notable exception is that if the new buffer allocation fails, it does so silently.  The process will then crash the next time someone tries to use the array, which will result in a null reference exception.

This function should assert or otherwise throw an error if its allocation call fails.

Janiels:
Fix #170 

Initially I wanted to do better error checking and handle the allocation failures
by returning errors from the APIs. However, there are very many usages and
since we do not use exceptions in the core this approach significantly worsens the code.

Since the `rcIntArray` usages are generally for small arrays I think just an assertion is fine,
while we can let the Recast fail if the bigger allocations fail.

### DIFF
--- a/Recast/Source/RecastAlloc.cpp
+++ b/Recast/Source/RecastAlloc.cpp
@@ -19,6 +19,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include "RecastAlloc.h"
+#include "RecastAssert.h"
 
 static void *rcAllocDefault(size_t size, rcAllocHint)
 {
@@ -77,6 +78,7 @@ void rcIntArray::doResize(int n)
 	if (!m_cap) m_cap = n;
 	while (m_cap < n) m_cap *= 2;
 	int* newData = (int*)rcAlloc(m_cap*sizeof(int), RC_ALLOC_TEMP);
+	rcAssert(newData);
 	if (m_size && newData) memcpy(newData, m_data, m_size*sizeof(int));
 	rcFree(m_data);
 	m_data = newData;


hymerman:
I'm happy with this - it's definitely better than before :)

Recast/Source/RecastAlloc.cpp
```diff
@@ -77,6 +78,7 @@ void rcIntArray::doResize(int n)
 	if (!m_cap) m_cap = n;
 	while (m_cap < n) m_cap *= 2;
 	int* newData = (int*)rcAlloc(m_cap*sizeof(int), RC_ALLOC_TEMP);
+	rcAssert(newData);
 	if (m_size && newData) memcpy(newData, m_data, m_size*sizeof(int));
```
jackpoz: what's the point of checking if newData is not NULL after rcAssert'ing it ?

grahamboree: If we run out of memory in a non-debug build, we don't want things to crash.  Maybe we should indicate that we ran out of memory somewhere when this happens, rather than failing silently?

Janiels: It will still crash later when the array is used, so that check could probably be eliminated.
I don't really see what we can do here - the whole problem is that, to detect that it has happened, we need to return an error code, and that needs to be bubbled up everywhere.
What we could do instead is use `std::terminate` or `std::abort` from inside this function, but I don't really think there's much point in it - I think it is very unlikely someone is meshing in environments where allocations in `rcIntArray` will fail (as said previously the allocations by it are very small usually).

PR #222, opened 2016-07-27
Fix findPath to return DT_OUT_OF_NODES correctly

### LINKED ISSUE #221: FindPath no longer returns the DT_OUT_OF_NODES detail flag
This is happening after my changes in #211 because it overrides the status unconditionally with the result from `getPathToNode`. Working on a fix.

Janiels:
findPath was unconditionally overriding the status it used to indicate
DT_OUT_OF_NODES for later. Restore this behavior so it properly can
return this failure flag.

Fix #221


hymerman:
Oh whoops. Not sure how we missed this! Good fix :+1:

PR #223, opened 2016-07-31
Fix a small typo in an error check

Janiels:
Just a tiny fix - wrong `dtStatus` variable was used here.

### DIFF
--- a/Detour/Source/DetourNavMeshQuery.cpp
+++ b/Detour/Source/DetourNavMeshQuery.cpp
@@ -481,9 +481,9 @@ dtStatus dtNavMeshQuery::findRandomPointAroundCircle(dtPolyRef startRef, const f
 	dtRandomPointInConvexPoly(verts, randomPoly->vertCount, areas, s, t, pt);
 	
 	float h = 0.0f;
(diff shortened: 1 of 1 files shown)


Detour/Source/DetourNavMeshQuery.cpp
```diff
@@ -481,9 +481,9 @@ dtStatus dtNavMeshQuery::findRandomPointAroundCircle(dtPolyRef startRef, const f
 	dtRandomPointInConvexPoly(verts, randomPoly->vertCount, areas, s, t, pt);
 	
 	float h = 0.0f;
-	dtStatus stat = getPolyHeight(randomPolyRef, pt, &h);
+	status = getPolyHeight(randomPolyRef, pt, &h);
```
grahamboree: Wouldn't this be overwriting any value set on line 429?  Maybe it should be `status |= getPolyHeight(randomPolyRef, pt, &h);`?

Janiels: You are right. I will write it the same way I did for `findPath`.
Currently on vacation though, will be back in a couple of weeks.

Janiels:
Closing until I can update this with feedback from @grahamboree.

PR #228, opened 2016-09-01
Properly use detail mesh from closestPointOnPoly

### LINKED ISSUE #227: Detour does not properly take detail mesh into account in closestPointOnPoly
When the point is not contained in the polygon Detour will not use the detail mesh properly.

Janiels:
When a point is outside the polygon being queried in
dtNavMesh::closestPointOnPoly and dtNavMeshQuery::closestPoinOnPoly,
Detour will interpolate along one of the edges to find the closest point
on the poly. However, Detour did not properly use the interpolated point
to find the correct height according to the detail mesh.

Fix #227


hymerman:
LGTM.

I did some more changes to this a few years back since even with this change it's still not quite doing an actual "closest point" in a geometric sense, which can create discontinuities if you're calling this regularly for a moving point, but this is definitely a step in the right direction! :+1:

Janiels:
I thought about that but I figured the epsilon was low enough that it shouldn't cause any issue. I might take a look in the future though.

PR #231, opened 2016-09-08
Take detail mesh into account for creating BV-tree

### LINKED ISSUE #230: Detour does not take detail mesh into account for the BV-tree
See comment here: https://github.com/recastnavigation/recastnavigation/blob/64385e9ed0822427bca5814d03a3f4c4d7a6db9f/Detour/Source/DetourNavMeshBuilder.cpp#L598

This causes problems for U shaped terrain when corner vertices are only created on the upper parts, or for A shaped terrain when corner vertices are only created on the lower parts. In these cases the polygon can extrude the bounding box used in the BV-tree significantly. This may cause functions that depend on querying polygons, like `queryPolygons` and `findNearestPoly`, to fail unexpectedly.

I have a fix that I am testing.

Janiels:
In U and A shaped terrain it was possible that polygons extruded their
BV-tree nodes significantly. This happened because all points for the
polygon (when excluding the detail mesh) were either on the high-level
terrain, or on the low-level terrain. This fixes the long-standing todo
about this.

As a result of this change, findNearestPoly and queryPolygons should
return more accurate results.

Fix #230

Note that Mikko talks about using "byte per bbox extents" in the todo.
I am not quite sure I understand that comment - can you explain what you
meant @memononen?

### DIFF
--- a/Detour/Source/DetourNavMeshBuilder.cpp
+++ b/Detour/Source/DetourNavMeshBuilder.cpp
@@ -168,45 +168,72 @@ static void subdivide(BVItem* items, int nitems, int imin, int imax, int& curNod
 	}
 }
 
-static int createBVTree(const unsigned short* verts, const int /*nverts*/,
-						const unsigned short* polys, const int npolys, const int nvp,
-						const float cs, const float ch,
-						const int /*nnodes*/, dtBVNode* nodes)
+static int createBVTree(dtNavMeshCreateParams* params, dtBVNode* nodes, int /*nnodes*/)
 {
 	// Build tree
-	BVItem* items = (BVItem*)dtAlloc(sizeof(BVItem)*npolys, DT_ALLOC_TEMP);
-	for (int i = 0; i < npolys; i++)
+	float quantFactor = 1 / params->cs;
+	BVItem* items = (BVItem*)dtAlloc(sizeof(BVItem)*params->polyCount, DT_ALLOC_TEMP);
+	for (int i = 0; i < params->polyCount; i++)
 	{
 		BVItem& it = items[i];
 		it.i = i;
-		// Calc polygon bounds.
-		const unsigned short* p = &polys[i*nvp*2];
-		it.bmin[0] = it.bmax[0] = verts[p[0]*3+0];
-		it.bmin[1] = it.bmax[1] = verts[p[0]*3+1];
-		it.bmin[2] = it.bmax[2] = verts[p[0]*3+2];
-		
-		for (int j = 1; j < nvp; ++j)
+		// Calc polygon bounds. Use detail meshes if available.
+		if (params->detailMeshes)
 		{
-			if (p[j] == MESH_NULL_IDX) break;
-			unsigned short x = verts[p[j]*3+0];
-			unsigned short y = verts[p[j]*3+1];
-			unsigned short z = verts[p[j]*3+2];
-			
-			if (x < it.bmin[0]) it.bmin[0] = x;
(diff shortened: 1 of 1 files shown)


memononen:
Good job! I once read an article about quantized AABB trees where the extents were stored as fraction of parent AABB (as bytes). I thought that'd be a good idea to implement. But later I decided to use the local quantized coodrinates (voxel corners) to do that instead. So you can ignore the comment.

Janiels:
@memononen Thanks Mikko, that sounds like an interesting idea. Do you by chance remember the name of the article?

I also thought about an even better solution to this problem we have here. The bounds may still be wrong (according to the terrain, by up to approximately the detail max error specified). What about keeping track of the minimum and maximum heights of the spans the polygon was created from through the meshing process, and pass this when creating nav mesh? It would be even more precise.

memononen:
I think it was one of the game programming gems books. I have a vague memory that it would save like <10% of memory but traversal was a lot slower due to int/float conversion and extra math.

Your suggestion a bout y-range sounds like a good idea to test. Another option would be to expand the y-range of the bounds by max detail mesh error.

Janiels:
@memononen Thanks for the link! I'll check the book out (if I recall correctly they actually have it at my university's library).

Regarding the detail mesh expanding: that would be a simple solution, but I am also concerned with correctness when the detail mesh is not built. I will do some experiments with them.

memononen:
@Janiels when you're not using detail mesh you have said good bye to height correctness already :) In order to get accurate polygon height extents you need to do the craziness of `getHeightData()`. I expect that when you're not using detail mesh it's because of generation speed. In that case the tile is probably small too, which alleviates things.

That being said, I think it is good that you give the problems that occur in productions a good hard thought. I think it might be worth thinking if you can calculate some guarantees to the height bounds. And it sounds like that's what you're doing already.

While I was working on the layers (the stuff that the tile cache is based on), I saw a lot of potential in the layered 2D representation. The `getHeightData()` has always been a source of frustration, and I really liked how the 2D representation makes many of those reverse mapping ambiguities go away. I was willing to would go as far as replace rcCompactHeightfield with that. Food for thought :)

hymerman:
> Perhaps people have just been solving it by increasing Y extents.

Of course :) Easy win! I remember doing exactly this 'fix' in Alien: Isolation when the alien started forgetting where he was when running over a bridge-shaped thing added late in development.

Good change - thanks!

Janiels:
@memononen That's a good point (about the height correctness). I definitely do not want to do the crazyness of `getHeightData` -- I was hoping the poly bounds could be passed along from somewhere so it would be fast too (eg. region creation, but I am not familiar enough with the area).

It just feels bad that the mesh is simplified so much that it becomes less usable/reliable than it ought to be. But since this has been the behavior for the past 5 years with minimal reports, and since it is even more mitigated by small tile sizes in the common case of no detail mesh (i.e. tile cache), I don't think I will do anything about it.

memononen:
@Janiels Don't feel too bad about that :) It is representation problem. The layer approach and monotone partitoning makes the two representations closer together, at the cost of worse polygonization. Ideally the process should be one way stream. On each step we transform the data to the next steps. Detail mesh breaks that deal and thus the pain.

One option could be to kill the details mesh and add more detail to the navmesh based on some height heuristic. I played around with this idea, and I had some heighfield filters to split things apart before region generation. I chose not to use that since the paths were less straight that way. This was before any replanning and line of sight optimizations I built for the crowd code.

PR #244, opened 2017-01-10
Add toggles for heightfield filtering

twoscomplement:
I wanted to see the effect of the various filtering options available, so I added toggles for each of them.

### DIFF
--- a/RecastDemo/Source/Sample.cpp
+++ b/RecastDemo/Source/Sample.cpp
@@ -42,7 +42,10 @@ Sample::Sample() :
 	m_crowd(0),
 	m_navMeshDrawFlags(DU_DRAWNAVMESH_OFFMESHCONS|DU_DRAWNAVMESH_CLOSEDLIST),
 	m_tool(0),
-	m_ctx(0)
+	m_ctx(0),
+	m_filterLowHangingObstacles(true),
+	m_filterLedgeSpans(true),
+	m_filterWalkableLowHeightSpans(true)
 {
 	resetCommonSettings();
 	m_navQuery = dtAllocNavMeshQuery();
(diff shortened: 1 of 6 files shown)


twoscomplement:
OK, I've added that.

RecastDemo/Source/Sample.cpp
```diff
@@ -42,7 +42,10 @@ Sample::Sample() :
 	m_crowd(0),
 	m_navMeshDrawFlags(DU_DRAWNAVMESH_OFFMESHCONS|DU_DRAWNAVMESH_CLOSEDLIST),
 	m_tool(0),
-	m_ctx(0)
+	m_ctx(0),
+	m_filterLowHangingObstacles(true),
```
hymerman: Nitpick: these should go between `m_navMeshDrawFlags` and `m_tool` to match member ordering in the class.

twoscomplement:
How about you add -Werror=reorder so this kind of thing is caught by the automated build, and I'll make that change.

hymerman:
Weird, I'm trying to, but can't get it to work locally. I added the flag to buildoptions in the Premake config but it's ignoring it. I can set it in CXXFLAGS in the Travis build script but I'd rather have it as part of the Premake config so it affects users before pushing. I'll try setting the 'FatalWarnings' flag for just Linux, as there are tons of warnings on other platforms right now - there are still a few warnings to fix up before pushing that though.

I figure that having warnings as errors on one platform is a good rabbit hole to have fallen down ;)

hymerman:
Done! Just need #245 and #248 to be merged, then this will start failing the Travis build.

hymerman:
Done - it doesn't look like there's a way to tell this PR to re-run the checks 'as if' it were already merged into master though, so they won't fail yet. Could you rebase and make the change please? I'd do it myself but I think it'll take the credit away from you :)

jakobbotsch:
Looks like closing and reopening worked.

hymerman:
Oh cool! You learn something new every day :) Thanks Jakob!

twoscomplement:
Nice! Thanks for making that change to the build.

I've changed the order of the members in the initialization list.

hymerman:
Looking good! Thanks again :)

PR #247, opened 2017-01-12
Upgrade Catch to 1.6.0

hymerman:
Currently there are loads of warnings in Catch code on Travis. This kind of thing:

    ../../../Tests/Recast/Tests_Recast.cpp: In function ‘void ____C_A_T_C_H____T_E_S_T____0()’:
    ../../../Tests/Recast/Tests_Recast.cpp:12:3: warning: suggest parentheses around comparison in operand of ‘==’ [-Wparentheses]
    ../../../Tests/Recast/Tests_Recast.cpp:13:3: warning: suggest parentheses around comparison in operand of ‘==’ [-Wparentheses]

I wondered if upgrading would fix them, but apparently not - no harm in taking the update anyway though I guess?

### DIFF
--- a/Tests/catch.hpp
+++ b/Tests/catch.hpp
@@ -1,6 +1,6 @@
 /*
- *  Catch v1.3.2
- *  Generated: 2015-12-28 15:07:07.166291
+ *  Catch v1.6.0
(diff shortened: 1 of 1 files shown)


jakobbotsch:
(Y)

PR #248, opened 2017-01-12
Set warnings as errors on Linux

hymerman:
Once #245 is merged, this should pass. It'll be nice to have warnings as errors for one CI platform!

It'll be much harder to do with AppVeyor since on Windows we get a ton of warnings about exception handling. Not sure how to tackle that for now.

### DIFF
--- a/RecastDemo/premake5.lua
+++ b/RecastDemo/premake5.lua
@@ -37,6 +37,12 @@ solution "recastnavigation"
 	configuration "windows"
 		defines { "WIN32", "_WINDOWS", "_CRT_SECURE_NO_WARNINGS" }
 
+	-- linux specific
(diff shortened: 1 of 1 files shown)


jakobbotsch:
I merged #245 but I guess we want to kick off the Travis CI again before merging this?
Do you know how to do that?

hymerman:
Haha, I'm just looking at how to do it now ;)

jakobbotsch:
Looks good 👍 
I'll let you merge.

hymerman:
Brilliant, thanks :)

FYI, I just rebased and forced a push - doesn't look like there's a way to do this from the web UI.

Also I really love that there's now an option to rebase instead of always merging! I didn't notice GitHub sneak that in!

PR #249, opened 2017-01-13
Suppress warnings when compiling for VS2015

twoscomplement:
The change to fastlz.c could also be implemented via "/wd4244" appropriately placed in premake5.lua

### DIFF
--- a/RecastDemo/Contrib/fastlz/fastlz.c
+++ b/RecastDemo/Contrib/fastlz/fastlz.c
@@ -81,6 +81,11 @@ typedef unsigned char  flzuint8;
 typedef unsigned short flzuint16;
 typedef unsigned int   flzuint32;
 
+/* Disable "conversion from A to B, possible loss of data" warning when using MSVC */
+#if defined(_MSC_VER)
+#pragma warning(disable: 4244)
+#endif
+
 /* prototypes */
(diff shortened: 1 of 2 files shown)


grahamboree:
I'm a bit reluctant to disable warnings globally like this because many people may be compiling us into a much larger codebase.  Maybe some `#pragma warning(push)` and `#pragma warning(pop)` lines would be good here.

hymerman:
The `#pragma` is ok because it's in the c file though :)

It looks like this removes some warnings but introduces some others - some stuff about DLLs in vs2010 and some unreachable code in vs2013. All is good in vs2015 though! Thanks! (except the stupid MSBuild warnings that the AppVeyor environment introduces)

@twoscomplement do you have those installed to take a look?

I'm inclined to accept this anyway as those particular warnings have been annoying me for a while :D

jakobbotsch:
Yes, this LGTM too. Sorry for slow review(s) @twoscomplement!

twoscomplement:
I don't have easy access to vs2010 or 2013 to look at those warnings, sorry.

jakobbotsch:
I looked at those warnings and it appears they are only in VS2010, and turning them off (with another definition, `_STATIC_CPPLIB`) will make warnings appear in newer versions. So I think it is ok to accept this as it is.

PR #250, opened 2017-01-13
Assertion failure custom function

domalb:
Implemented the ability to set custom function to be called in case of assertion failure.

This functionality is similar to custom allocation functions.

### DIFF
--- a/Detour/Include/DetourAssert.h
+++ b/Detour/Include/DetourAssert.h
@@ -23,11 +23,34 @@
 // Feel free to change the file and include your own implementation instead.
 
 #ifdef NDEBUG
+
 // From http://cnicholson.net/2009/02/stupid-c-tricks-adventures-in-assert/
 #	define dtAssert(x) do { (void)sizeof(x); } while((void)(__LINE__==-1),false)  
+
 #else
+
+/// An assertion failure function.
+//  @param[in]		expression  asserted expression.
+//  @param[in]		file  Filename of the failed assertion.
+//  @param[in]		line  Line number of the failed assertion.
+///  @see dtAssertFailSetCustom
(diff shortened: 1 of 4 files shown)


hymerman:
Hi @domalb, thanks for contributing! This is great stuff and is something that's definitely needed.

Your implementation is in keeping with the style of the project and will have minimal impact on existing users, so that's cool.

This is something I've been planning to do for a while, and I've got a commit of my favourite assertion macro implementation (copied from other personal projects) in a WIP branch here (not committed yet as it's part of a larger change to add some better unit tests of code that fails asserts): https://github.com/hymerman/recastnavigation/commit/e6980dddb286f191c38f13e4c313c7f6ac822634

@jakobbotsch and @grahamboree do you think it's worth merging this now or should I have a go at tidying up my one and merging? I'm biased towards code I write so I won't make the call ;)

domalb:
@hymerman would you prefer that I update the pull request from now to prepare for your future commit and reduce api changes ?
- an assertion failure response halt/continue
- a message with variadic arguments for formatting

jakobbotsch:
I think we should accept this as it is. Your changes @hymerman seem a bit more pervasive and not as much in line with standard Recast coding style (C-style, like it or not - but it's good to be consistent).

jakobbotsch:
Sounds good @hymerman?

hymerman:
Thanks for the prod - forgot about this!

Ok, I've had a think and reckon we should just take this for now. Thanks @domalb :)

I'd love to get a tidied up version of mine in for a larger release though, as it's had a lot of thought go into it and does just drop in, and it's actually just free functions and macros so does still architecturally fit Recast style. It's just a case of finding free time, whatever that is...

jakobbotsch:
Thanks for the contribution!

PR #252, opened 2017-01-16
Added new agent state and custon steering types.

sagaceilo:
This PR adds new agent state when its still collider but cannot move. Another agents also cannot move disabled agent. This way is easy to create dynamic small obstacles w/o actually rebaking or using cache.

Steering has new state that uses angular rotation instead vector interpolation. It helps to manage agents that require limited speed of rotation.


sagaceilo:
Sorry, something went wrong on Windows GitHub :/

PR #253, opened 2017-01-25
Debug draw duPolyToCol custom function

domalb:
Added the ability for the user to set a custom function that computes the color of a polygon when debug-drawing a mesh.

This way, the user can get a debug draw that fits his particular situation.
He can scan the polygon (mainly flags & area) and determine the color that expresses it best (e.g. green for grass flag, blue for water flag, mud for grass & water flags...).

Default behavior is kept unchanged. 

Possible further improvement : color constants defined on the fly in function implementations could be actual global variables that the user could set.

### DIFF
--- a/DebugUtils/Include/DebugDraw.h
+++ b/DebugUtils/Include/DebugDraw.h
@@ -69,9 +69,8 @@ struct duDebugDraw
 
 	/// Compute a color for given polygon.
 	/// @param poly [in] the polygon to compute color for.
-	/// @param alpha [in] alpha component that should be set in computed color.
 	/// @return The color value.
-	virtual unsigned int polyToCol(const struct dtPoly* poly, int alpha);
+	virtual unsigned int polyToCol(const struct dtPoly* poly);
 };
 
 inline unsigned int duRGBA(int r, int g, int b, int a)
--- a/DebugUtils/Include/DebugDraw.h
+++ b/DebugUtils/Include/DebugDraw.h
@@ -66,6 +66,12 @@ struct duDebugDraw
 	
 	/// End drawing primitives.
 	virtual void end() = 0;
+
+	/// Compute a color for given polygon.
+	/// @param poly [in] the polygon to compute color for.
+	/// @param alpha [in] alpha component that should be set in computed color.
+	/// @return The color value.
+	virtual unsigned int polyToCol(const struct dtPoly* poly, int alpha);
 };
 
 inline unsigned int duRGBA(int r, int g, int b, int a)
--- a/DebugUtils/Include/DetourDebugDraw.h
+++ b/DebugUtils/Include/DetourDebugDraw.h
@@ -30,15 +30,6 @@ enum DrawNavMeshFlags
 	DU_DRAWNAVMESH_COLOR_TILES = 0x04,
 };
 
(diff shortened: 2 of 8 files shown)


jakobbotsch:
Would it make more sense to add this as a virtual method to `duDebugDraw`?

domalb:
Yes, that feature could be implemented as a virtual method of duDebugDraw.
The change is straightforward.
It would make it less C like and more C++ like, which seems good to me.

But I previously noted two points against that:
1) duDebugDraw currently is a pure abstract interface, whereas this new method will have a default implementation. Should duDebugDraw be kept abstract for any reason?
2) As far as I can see, duDebugDraw only deals with applying draw commands to rendering implementation (GL, DX, or any plugged engine) using native types.
duDebugDraw does not deal with higher level detour-related structures/concepts.
Adding a polyToCol method sounds a bit out of its role, to me. 

What do you think?

jakobbotsch:
1 is no problem, virtual method with a default implementation is ok. For 2 it was something I thought about too, but `duDebugDraw` is named in such a way that I think it makes sense to have R&D-specifics in it.
My main point of gripe with the current implementation in this PR is the inconsistency in the methods used to allow user-customization in the same file (with the callback vs virtual).

domalb:
Allocation and assertion user customizations could be easily changed to this interface principle.
Did I miss another user-customization feature already in place ?

For good consistency, it would be better to merge with PolyToCol implemented with callback and move all user customizations to virtuals at once.
I could handle this later.

jakobbotsch:
Assertion and allocation as static callbacks is more in line with other libraries I have seen. Also, to make them virtual we would still need somewhere to put them and the object.

I don't think you've missed anything, I just think in this case there is a good place to put it (`duDebugDraw`) already. @hymerman @grahamboree what do you think?

hymerman:
I agree - `duDebugDraw` is a good fit for this, even if it is a bit inconsistent with other customisation points (though it's already that way so never mind!)

domalb:
Done

jakobbotsch:
Almost looks good to me, I am just wondering whether the alpha is necessary to pass to the virtual method, or if we should manually call `duTransCol` instead. Do you need to modify the alpha in your use-case?

domalb:
In my use case I did not bother about alpha, `duTransCol` would be OK.

Maybe using something like `duTransCol` but which multiplies the alpha instead of overwriting it would be even better ? That way the user has control over alpha nuances, but the overall alpha (currently a constant, but could be a variable in the future) still applies.

jakobbotsch:
I am sorry for overcomplicating this but I just don't think the alpha fits well as a param and it is hard to change in the future. So I would go for removing the arg and just calling `duTransCol` for now. A multiplication can be added, I wouldn't really have a problem with that, but it does have a limitation: if we multiply by eg. `* 0.25` a user is still not going to be able to customize the alpha to access the upper 3/4th of values.

domalb:
No problem, I like to quibble. There is always something to learn.

Regarding to the `0.25` example, you are right with the limitation. That's the constant I talked about above. In the future that constant may become a variable the user could choose (whatever the mean), so the user can have access to the full alpha range.

jakobbotsch:
Thanks!

PR #255, opened 2017-02-01
PolyToCol changed to AreaToCol

domalb:
@jakobbotsch I followed your Git CLI tutorial, and it worked, great !
I'm still a bit confused by the need of using command line instead of GUI, despite the numerous released clients and my very basic needs.

Anyway, the area to col is applied more widely in the demo, which is good.


jakobbotsch:
Thanks!

PR #256, opened 2017-02-02
Change back to previous default color

jakobbotsch:
I was a bit too fast to merge #254 before checking how the colors actually look.
Turns out the new default color changed significantly; change it back to the old original to retain that look.


jakobbotsch:
I have also changed the road color since it was a bit hard to differentiate to the input geometry color.

domalb:
No problem, that's ok.

PR #258, opened 2017-02-04
RecastDemo: add Load & Save buttons for SoloMesh Sample

aurelien-rainone:
### Add support for loading and saving a solo mesh
 - Solo mesh sample is serialized/unserialized with the same format as a **tiled mesh**. It's just a tiled mesh with an unique tile. That means a solo mesh binary is compatible is a tiled mesh loader.
 - NavMesh is saved to/loaded from `solo_navmesh.bin` binary file.
 - Add and connect the corresponding 'Save/Load' GUI elements.

I took the liberty to not speak with anybody here first about this feature because I already had it implemented. I'm going forward with my [Go port of Detour and Recast](https://github.com/aurelien-rainone/go-detour), that does not support tiled mesh for now. As part of the test suite I was needing a way to save and load solo meshes from the gui, for I'm comparing the binaries produced in Go with the ones from the original library.

### DIFF
--- a/RecastDemo/Source/Sample_SoloMesh.cpp
+++ b/RecastDemo/Source/Sample_SoloMesh.cpp
@@ -79,20 +79,164 @@ void Sample_SoloMesh::cleanup()
 	dtFreeNavMesh(m_navMesh);
 	m_navMesh = 0;
 }
-			
+
+static const int NAVMESHSET_MAGIC = 'M'<<24 | 'S'<<16 | 'E'<<8 | 'T'; //'MSET';
+static const int NAVMESHSET_VERSION = 1;
+
+struct NavMeshSetHeader
+{
+	int magic;
+	int version;
+	int numTiles;
+	dtNavMeshParams params;
+};
+
+struct NavMeshTileHeader
+{
+	dtTileRef tileRef;
+	int dataSize;
+};
+
+void Sample_SoloMesh::saveAll(const char* path, const dtNavMesh* mesh)
+{
+	if (!mesh) return;
+
+	FILE* fp = fopen(path, "wb");
+	if (!fp)
+		return;
+
+	// Store header.
+	NavMeshSetHeader header;
+	header.magic = NAVMESHSET_MAGIC;
+	header.version = NAVMESHSET_VERSION;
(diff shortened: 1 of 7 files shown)


RecastDemo/Source/Sample_SoloMesh.cpp
```diff
@@ -79,20 +79,164 @@ void Sample_SoloMesh::cleanup()
...
+		if (!tileHeader.tileRef || !tileHeader.dataSize)
+			break;
+
+		unsigned char* data = (unsigned char*)dtAlloc(tileHeader.dataSize, DT_ALLOC_PERM);
+		if (!data) break;
+		memset(data, 0, tileHeader.dataSize);
+		readLen = fread(data, tileHeader.dataSize, 1, fp);
+		if (readLen != 1)
+		{
+			fclose(fp);
```
jakobbotsch: `data` is leaked here

jakobbotsch:
Hi,

I am so sorry for being this late with a review. I had a small request for a change, otherwise it looks good.

RecastDemo/Source/Sample_SoloMesh.cpp
jakobbotsch: I notice now that this code is copy pasted from `Sample_TileMesh::loadAll` which has the same bug. It would be nice to fix that bug and move the common code into the `Sample` class instead.

aurelien-rainone:
I fixed the leak and moved the common code in the `Sample` class. In order to build `RecastDemo` on gcc7/linux without modifying the premake script, I had to fix the _implicit-fallthrough_ warnings (more info in commit 0cd93d4 message). Anyway I think this warning is reasonable and the comment enforce code comprehension.

Before merging, please try to run `RecastDemo` and see if it works properly (also saving and loading navmeshes), unfortunately I couldn't do it myself, I updated my system very recently and it seems something is broken with X11/OpenGL (though other GL applications work...). I couldn't figure out a solution tonight (it's late...:-), it might be due to an error in the last 2 commits for example or something completley different but I prefer to warn you.

jakobbotsch:
Thank you, finally I was able to test and get this in. :)

PR #271, opened 2017-05-14
Fix 'dtRandomPointInConvexPoly()' returns a garbage value if s == 1.0f.

gasbank:


hymerman:
Fix looks good to me - good catch! And thanks for adding a unit test. I checked the test without the fix and it fails correctly, and gave a few other tests a quick go which work as expected.

I'll leave this open for the other maintainers to check over before accepting.

jakobbotsch:
LGTM too. Good catch indeed.

PR #273, opened 2017-06-27
Add RecastConfig.h and DetourConfig.h

jakobbotsch:
I am starting some improvements on R&D. To begin with, I want to do some of the cleanup that @hymerman has suggested earlier by moving configs into their own files.
If I missed any config types/variables/values please let me know!


hymerman:
Looks good - good doc tweaks too. I'll let you merge when you can :)

I haven't got time to check for other config things that can go in there but that's something that can be done over time!

Issue #274, opened 2017-06-28
Unification of ability flags and area types


jakobbotsch:
Currently we support 64 area types and 16 ability flags. I have had some gripes with this split when using R&D:
1. Often we want to mark areas with flags too. For example, there might be a city that is cheap to navigate for one kind of agent, but expensive for a different kind. If we have water, ground, road in this city then we would ideally like to mark the whole city with an additional flag, while still retaining the water/ground/road types. But this quickly reduces the amount of area types we have available.
1. 16 ability flags is, in my experience, more than enough. In fact we would often like to trade some of these for more area types.
1. Marking polygons with the correct ability flags is slightly hard because only area types are passed through the entire meshing process. See eg. https://github.com/recastnavigation/recastnavigation/blob/46654531e4e5c17d2ebe1933f3e768f908540f03/RecastDemo/Source/Sample_TileMesh.cpp#L1227 for how we do this right now, but I have had issues with being able to differentiate between different "Ground" polygons to give them correct ability flags.
1. In addition, the split might cause confusion for newer users (I know that I was confused in the beginning!)

I believe we should unify these two concepts to a single, larger variable. I also belive that @memononen did this for Unity's implementation of R&D. Coupled with #260 this will in general give better flexibility in the meshing process.

Now there are a few things to take into account:
1. This will have to be a breaking change, so we will have to bump the mesh version. I don't expect this to be any issue as this will go into 2.0.
1. `dtQueryFilter` will probably have to be made virtual in all cases. I am not quite sure how to write a generic version of it that supports extracting flags/types and calculating costs in a general way. @memononen how was this done in Unity?

I will be starting the work on this very soon.

memononen:
Before you start working on this, I recommend to take a look at how we got there:
http://digestingduck.blogspot.fi/2009/09/area-progress.html
http://digestingduck.blogspot.fi/2009/10/status-update.html
http://digestingduck.blogspot.fi/2010/01/fresh-start-for-areas.html
http://digestingduck.blogspot.fi/2010/02/area-progress-pt-4.html
http://digestingduck.blogspot.fi/2010/02/area-progress-pt-5.html

The area=ability was a bad choice at Unity, done before my time there. That being said, the current situation is not perfect either.

I think it is very important to realise that there are two loosely connected ideas there: 1) travel cost, 2) travel capability. I think these are well reflected in Detour (even if the data type is small), but the build process handles them poorly.

[...]

PR #275, opened 2017-06-28
[WIP] Unify area types and ability flags

jakobbotsch:
This is work-in-progress for #274.


jakobbotsch:
This turned out to be a bit more work/pervasive than I expected it to be. For now I will close this out and probably try to come back later when I have more time.

PR #279, opened 2017-07-25
Renamed all instances of 'extents' to 'halfExtents'

aymarfisherman:
And left one duplicate method to maintain backwards compatibility.

### DIFF
--- a/DetourCrowd/Source/DetourCrowd.cpp
+++ b/DetourCrowd/Source/DetourCrowd.cpp
@@ -386,7 +386,8 @@ bool dtCrowd::init(const int maxAgents, const float maxAgentRadius, dtNavMesh* n
 	m_maxAgents = maxAgents;
 	m_maxAgentRadius = maxAgentRadius;
 
-	dtVset(m_halfExtents, m_maxAgentRadius*2.0f,m_maxAgentRadius*1.5f,m_maxAgentRadius*2.0f);
+	// Larger than agent radius because it is also used for agent recovery.
+	dtVset(m_agentPlacementHalfExtents, m_maxAgentRadius*2.0f, m_maxAgentRadius*1.5f, m_maxAgentRadius*2.0f);
 	
 	m_grid = dtAllocProximityGrid();
 	if (!m_grid)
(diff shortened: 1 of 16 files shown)


aymarfisherman:
Here you go.
I left one method duplicated: getQueryExtents and getQueryHalfExtents to provide backwards compatibility.
If you spot anything you'd like changed, just let me know.
Cheers.

DetourCrowd/Source/DetourCrowd.cpp
```diff
@@ -386,7 +386,7 @@ bool dtCrowd::init(const int maxAgents, const float maxAgentRadius, dtNavMesh* n
 	m_maxAgents = maxAgents;
 	m_maxAgentRadius = maxAgentRadius;
 
-	dtVset(m_ext, m_maxAgentRadius*2.0f,m_maxAgentRadius*1.5f,m_maxAgentRadius*2.0f);
+	dtVset(m_halfExtents, m_maxAgentRadius*2.0f,m_maxAgentRadius*1.5f,m_maxAgentRadius*2.0f);
```
jakobbotsch: Is this a (preexisting) bug? This ought to be only `m_maxAgentRadius` for X and Z, as far as I can see.
cc @recastnavigation/write

jakobbotsch:
Thank you! It looks good to me. I will let someone else mull it over before merging.
I hope the `getQueryExtents` and `getQueryHalfExtents` will not cause too many misunderstandings.

@grahamboree @hymerman PTAL as well

DetourCrowd/Source/DetourCrowd.cpp
memononen: It's intentional. Better name might be a good idea. If nothing else `m_agentPlacementHalfExtents`.

The intent is that it defines how far off the agent can be outside the navmesh to be "snapped" back to navmesh in case of shit-happens. Two reasons it is small. I wanted the agents to get stuck/stop when too far outside navigable area, and secondly large query extents is expensive to process.

jakobbotsch: I see, that makes sense. However it is a little nondeterministic since `findNearestPoly` can return polys that do not actually touch the extents (on account of `queryPolygons` using AABB's for the polys internally).

@aymarfisherman do you mind making the renaming suggested, now that you're renaming here already?

aymarfisherman: Not at all, may I suggest `m_agentRecoveryHalfExtents` ? Let me know what you guys think and then I'll do this renaming.

jakobbotsch: It is used during normal operation too, not only if it detects that it needs recovering, so that might be a little misleading. I think Mikko's suggestion, maybe with a comment on the field that it is chosen large for recovery reasons, is best.

aymarfisherman:
Done.

jakobbotsch:
Hi, sorry for forgetting about this! Since it looks good to me I have gone ahead and merged it.

PR #282, opened 2017-08-24
Return OOM error when navmeshtile allocations fail

mendsley:
This was previsously return a DT_SUCCESS mask, as the prior
call to dtAllocTileCache* had succeeded.


jakobbotsch:
Thank you for the contribution!

PR #287, opened 2017-11-01
Bump up RC_SPAN_HEIGHT_BITS.

mbabinski-at-google:
This does not affect the size of rcSpan on x64, and makes it less likely users will need to modify this.

### DIFF
--- a/Recast/Include/Recast.h
+++ b/Recast/Include/Recast.h
@@ -263,7 +263,7 @@ struct rcConfig
 };
 
 /// Defines the number of bits allocated to rcSpan::smin and rcSpan::smax.
-static const int RC_SPAN_HEIGHT_BITS = 13;
+static const int RC_SPAN_HEIGHT_BITS = 16;
 /// Defines the maximum value for rcSpan::smin and rcSpan::smax.
 static const int RC_SPAN_MAX_HEIGHT = (1 << RC_SPAN_HEIGHT_BITS) - 1;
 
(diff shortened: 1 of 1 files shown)


Recast/Include/Recast.h
```diff
@@ -277,7 +277,7 @@ struct rcSpan
 {
 	unsigned int smin : RC_SPAN_HEIGHT_BITS; ///< The lower limit of the span. [Limit: < #smax]
 	unsigned int smax : RC_SPAN_HEIGHT_BITS; ///< The upper limit of the span. [Limit: <= #RC_SPAN_MAX_HEIGHT]
-	unsigned int area : 6;                   ///< The area id assigned to the span.
+	unsigned int area : 8;                   ///< The area id assigned to the span.
```
jakobbotsch: There are a few more places that have some implicit dependence on this value, eg. `dtPoly::areaAndtype`. There might also be some other dependencies in the meshing process, though I do not have time to check closer at the moment. But I think you should leave this as is. See also #274

mbabinski-at-google: Reverted.

jakobbotsch:
Sorry for the wait, but I had a super busy week and I wanted to track down some usages in the build process.
After some consideration I have decided not to take this one, for a couple of reasons. For one there are still some internal caps on this value (which ought to be documented/commented): https://github.com/recastnavigation/recastnavigation/blob/master/Recast/Source/Recast.cpp#L409
If we want to increase it, I think we need to take care of this value too (or make it a regular 16 bit data type) -- otherwise the bit field is a bit misleading. However unfortunately increasing the size beyond 16 bits ends up being quite an extensive and breaking change. 

With this in mind I would prefer to leave it as is, since this is a 50% increase in the size on 32-bit with no performance improvement, and since it is pretty easy to customize for the user in case that is needed.

Apologies for wasting some of your time on this. :-(

PR #293, opened 2017-12-14
fixed load .gset file error

chenshungen:
The previous code can not read the .gset file, because of strlen(.gset) > 4 . when strlen(ext) > 4 strncmp(current->d_name + len - 4, ext.c_str(), 4) == 0 is aways false


jakobbotsch:
Thank you!

PR #297, opened 2017-12-31
allow 64 bit windows builds

jarrettchisholm:


jakobbotsch:
Thanks for the contribution, but the build config is only used for building the demo, and this does not work for the demo (requires changes to the SDL paths). For integrating the core libraries, we recommend to copy/paste the needed directories into your source tree, and then configuring x86/x64 in your own build system.

PR #301, opened 2018-01-06
Allow 64 bit windows builds

jarrettchisholm:

### DIFF
--- a/Recast/Include/RecastAlloc.h
+++ b/Recast/Include/RecastAlloc.h
@@ -308,7 +308,7 @@ public:
 		m_impl.pop_back();
 		return v;
 	}
-	int size() const { return m_impl.size(); }
+	int size() const { return static_cast<int>(m_impl.size()); }
 	int& operator[](int index) { return m_impl[index]; }
 	int operator[](int index) const { return m_impl[index]; }
 };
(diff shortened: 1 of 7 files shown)


jakobbotsch:
Hi, thank you for this. Sorry for getting to it so very late.

Would you mind updating the AppVeyor build to be 64-bit as well? I don't think we need a 32-bit build, so you can just change it. I believe you just need to add:

```yaml
platform:
  - x64
```

Below `configuration`.

jakobbotsch:
Looks like a couple of explicit casts or type updates are needed:
  c:\projects\recastnavigation\recastdemo\source\sample_tempobstacles.cpp(1342): warning C4267: '=': conversion from 'size_t' to 'int', possible loss of data [C:\projects\recastnavigation\RecastDemo\Build\vs2017\RecastDemo.vcxproj]
  c:\projects\recastnavigation\recastdemo\source\imgui.cpp(36): warning C4267: 'initializing': conversion from 'size_t' to 'unsigned int', possible loss of data [C:\projects\recastnavigation\RecastDemo\Build\vs2017\RecastDemo.vcxproj]

jarrettchisholm:
Should be fixed now.

jakobbotsch:
Thank you!

PR #303, opened 2018-01-09
Provide better feedback when dtNavMesh::addTile fails due to a tile already being present at a given x,y

xeropresence:
Failure because the tile coordinate is already occupied is the only failure that does not specify more information.


jakobbotsch:
Thanks!

PR #307, opened 2018-02-01
Don't discard changes that occur after hitting MAX_UPDATE

mastahg:
Currently if we hit the maximum number of requested changes, the rest of the changes are discarded. This pr changes it so that we always complete at least one obstacle, and the remaining obstacles are then moved to the front of the line to be processed next update.


DetourTileCache/Source/DetourTileCache.cpp
```diff
@@ -546,12 +549,13 @@ dtStatus dtTileCache::update(const float /*dt*/, dtNavMesh* navmesh,
...
-					{
-						if (!contains(m_update, m_nupdate, ob->touched[j]))
-							m_update[m_nupdate++] = ob->touched[j];
-						ob->pending[ob->npending++] = ob->touched[j];
-					}
+					if (m_nupdate == MAX_UPDATE && i != m_nreqs - 1)
+						abortedEarly = true;
+
+					if (!contains(m_update, m_nupdate, ob->touched[j]))
+						m_update[m_nupdate++] = ob->touched[j];
```
jakobbotsch: This can overflow `m_update` now.

DetourTileCache/Source/DetourTileCache.cpp
```diff
@@ -562,19 +566,38 @@ dtStatus dtTileCache::update(const float /*dt*/, dtNavMesh* navmesh,
...
```
jakobbotsch: Same as above.

PR #310, opened 2018-03-15
Support build by CMake

elsid:


psi29a:
This pretty much makes the project ready to be packaged downstream in Debian and Ubuntu. :)

psi29a:
Added support for versioned shared libraries that are platform agnostic. 
With this PR, we can move forward with inclusion in Debian/Ubuntu. :)

jakobbotsch:
Thanks for the contribution!
I am not too familiar with CMake. Is it normal to change all includes to angled-bracket style? I notice that other projects using CMake do not do this, for example LLVM. What is the reason for this change? (Ideally adding a build-system would not be invasive to our original files.)

elsid:
>Is it normal to change all includes to angled-bracket style?
It is standart practice when headers is installing into system directory. But it is not required. So I've removed all those changes.

monyarm:
So, is this going to be merged ?

jakobbotsch:
Hi, yes, super busy lately. I think it all looks fine to me, but as I said I have almost no experience with CMake, so let me tag some of the other maintainers as well: @grahamboree @hymerman. I expect they are pretty busy as well, so I will just merge it if they haven't dropped in in a couple days time.

hymerman:
Thanks all - this is a great contribution!

I agree totally on the direction to take with build tools, Jakob.

jakobbotsch:
@elsid @psi29a

As part of updating Travis and AppVeyor to make sure the CMake builds are running with warnings as errors I have hit a bit of a snag. I am unable to get CMake to build and run the demo Windows. There are a few issues:
1. We fail to find SDL2
2. On Windows, we should build with static libraries by default, as we get linker errors otherwise
3. When built, the supporting files (SDL2.dll, meshes, test cases) are not properly copied to the binary directory

My attempts to solve these can be seen here: https://github.com/jakobbotsch/recastnavigation/commit/5522402a762c179f2e74fb9ca35621f3e3179e31
I am not sure if this is the correct way to solve these, but it seems to fix 1 and 2. However, I don't know how to solve 3. CMake seems to make it overly complicated to copy something to the output directory, at least with Visual Studio...

Could you please advice on how to fix this correctly? Ideally we should get to the same state as where we are with premake5, i.e. the ability to just run premake and then use F5 to launch and debug with Visual Studio. Otherwise I am inclined to keep premake as the main build system and eventually remove Cmake.

jakobbotsch:
I have followed your steps and it still does not work.
```powershell
git clone https://github.com/elsid/recastnavigation.git
cd recastnavigation
git checkout ec5c5bf
# Downloaded and placed SDL in RecastDemo/Contrib/SDL directory
mkdir build
cd build
mkdir install
cmake -G "Visual Studio 15 2017 Win64" -D RECASTNAVIGATION_STATIC=ON -D SDL2_ROOT_DIR="..\RecastDemo\Contrib\SDL" -D CMAKE_INSTALL_PREFIX=install ..
cmake --build . --target install
```

Files are not automatically copied to the output directory. `install\bin\` contains only a single file, `RecastDemo.exe` which does not run (because of missing dependencies). Even if this worked, having to manually configure Visual Studio to set up debugging afterwards is a significant downside to switching to CMake. This all works out of the box with our Premake5 build.

elsid:
Try this commits:
https://github.com/elsid/recastnavigation/commit/696ae741c9204ab5643632f7a5f472cf8eadb721
https://github.com/elsid/recastnavigation/commit/935e2626982e3a7956d148afc26aad017a4a294d

If you want to run cmake from visual studio you have to configure it. I don't see how it could work otherwise.

jakobbotsch:
That works on install now (except the output subsystem is console instead of window), however it still does not work from Visual Studio.

> If you want to run cmake from visual studio you have to configure it. I don't see how it could work otherwise.

CMake and Premake both generate Visual Studio files (.sln, .vcxproj etc.). I want to be able to open the .sln file in Visual Studio, set RecastDemo as the startup project and run it under the debugger. I don't want to have manually set up VS to debug the RecastDemo.exe from the install directory and then have to run the install target manually every time I want to debug.

elsid:
@jakobbotsch 
Fix for console https://github.com/elsid/recastnavigation/commit/ca9c159ae61bf3e9882d50d04df86fc81f6840ee

And fix to be able run debug from generated project https://github.com/elsid/recastnavigation/commit/78a7137f2af7d96554c8ceef71c22a49fd9545e4

jakobbotsch:
Thank you, it all seems to work well now!
Can we default RECASTNAVIGATION_STATIC to ON on Windows and SDL2_ROOT_DIR to RecastDemo\Contrib\SDL? And finally, would you mind opening a PR with these fixes?

PR #311, opened 2018-03-15
Unconditionally initialize rcCompactHeightfield borderSize

elsid:
To avoid uninitialized value when borderSize <= 0.


jakobbotsch:
Thanks!

PR #313, opened 2018-04-25
Remove redundant copying in rcBuildRegions.

mbabinski-at-google:
rcBuildRegions uses dstReg and dstDist as scratch space. Currently, it copies src* -> dst* on each iteration of the inner loop of expandRegions. This is a costly operation; on a sampled map, it copies 16Mbytes of data, when only ~6kbytes are accessed. We modify rcBuildRegions and expandRegions to keep src* and dst* in sync, except during the main loop of expandRegions.

This has reduced runtime of the Recast portion of my build pipeline by 70% (!). memcpy() no longer appears in sampled profiles at all.

### DIFF
--- a/Recast/Include/RecastAlloc.h
+++ b/Recast/Include/RecastAlloc.h
@@ -20,6 +20,7 @@
 #define RECASTALLOC_H
 
 #include <stddef.h>
+#include <vector>
 
 /// Provides hint values to the memory allocator on how long the
 /// memory is expected to be used.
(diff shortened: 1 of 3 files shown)


jakobbotsch:
Wow, that is quite a significant reduction in your runtime.
The change looks good, but I was wondering if we can do even better. If we're going to store the dirty indices, can we just store the results in that array? I.e. we store `i`, `r` and `d2` and then write them into the `srcReg` and `srcDist` arrays directly after we are done. We should be able to get completely rid of `dstReg` and `dstDist` arrays and the swapping with that approach. Would you mind checking the results of this approach?

mbabinski-at-google:
Good call. Unfortunately, I botched up my repo, and screwed up this PR. Making a new one.

PR #314, opened 2018-04-27
Remove redundant copying in rcBuildRegions.

mbabinski-at-google:
rcBuildRegions uses dstReg and dstDist as scratch space. Currently, it copies src* -> dst* on each iteration of the inner loop of expandRegions. This is a costly operation; on a sampled map, it copies 16Mbytes of data, when only ~6kbytes are accessed. We get rid of dst*, and modify expandRegions to keep track of deltas and apply them at the end of each iteration.

This has reduced runtime of the Recast portion of my build pipeline by 70% (!). memcpy() no longer appears in sampled profiles at all.

### DIFF
--- a/Recast/Source/RecastRegion.cpp
+++ b/Recast/Source/RecastRegion.cpp
@@ -386,8 +386,6 @@ static unsigned short* expandRegions(int maxIter, unsigned short level,
 	}
 
 	rcIntArray dirtyEntries;
-	memcpy(dstReg, srcReg, sizeof(unsigned short)*chf.spanCount);
-	memcpy(dstDist, srcDist, sizeof(unsigned short)*chf.spanCount);
 	int iter = 0;
 	while (stack.size() > 0)
 	{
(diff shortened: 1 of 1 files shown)


Recast/Source/RecastRegion.cpp
```diff
@@ -427,18 +425,22 @@ static unsigned short* expandRegions(int maxIter, unsigned short level,
...
 		}
 		
-		// rcSwap source and dest.
-		rcSwap(srcReg, dstReg);
-		rcSwap(srcDist, dstDist);
+		// Copy entries that differ between src and dst to keep them in sync.
+		for (int i = 0; i < dirtyEntries.size(); i+=3) {
+			int idx = dirtyEntries[i];
+			srcReg[idx] = dirtyEntries[i+1];
+			srcDist[idx] = dirtyEntries[i+2];
```
jakobbotsch: We need casts here:
```
  c:\projects\recastnavigation\recast\source\recastregion.cpp(441): warning C4244: '=': conversion from 'int' to 'unsigned short', possible loss of data [C:\projects\recastnavigation\RecastDemo\Build\vs2015\Recast.vcxproj]
  c:\projects\recastnavigation\recast\source\recastregion.cpp(442): warning C4244: '=': conversion from 'int' to 'unsigned short', possible loss of data [C:\projects\recastnavigation\RecastDemo\Build\vs2015\Recast.vcxproj]
```

jakobbotsch:
Yes, I was a bit surprised that the Travis build succeeded, since I believe we pass both -Wpedantic and -Werror, although I can't seem to find the options in the logs in Travis. @hymerman do you have any idea why?

Anyway, thanks for the contribution!

hymerman:
I think we have -Wall and -Werror for linux/mac. Not sure why those don't catch it! I can't see much in the build logs really. It's a mystery :)

PR #318, opened 2018-05-02
Add homebrewed rcVector class and tests, and use it to implement rcIntArray.

mbabinski-at-google:
This implements most of the std::vector interface, with some minor elisions of
infrequently used methods such as "insert", and some STL bookkeeping.
 Performance is favorable compared to std::vector, and, in some cases, somewhat
better than naive code to manually manage a malloc'd dynamic array.

BM_FlatArray_Push:                     100 iterations in   72889721 nanos:  728897.21 nanos/it
BM_FlatArray_Fill:                     100 iterations in    2700746 nanos:   27007.46 nanos/it
BM_FlatArray_Memset:                   100 iterations in    2687810 nanos:   26878.10 nanos/it
BM_rcVector_Push:                      100 iterations in   51114316 nanos:  511143.16 nanos/it
BM_rcVector_PushPreallocated:          100 iterations in    9532106 nanos:   95321.06 nanos/it
BM_rcVector_Assign:                    100 iterations in    2371534 nanos:   23715.34 nanos/it
BM_rcVector_AssignIndices:             100 iterations in    2573052 nanos:   25730.52 nanos/it
BM_rcVector_Resize:                    100 iterations in    2455761 nanos:   24557.61 nanos/it
BM_stdvector_Push:                     100 iterations in   71739167 nanos:  717391.67 nanos/it
BM_stdvector_PushPreallocated:         100 iterations in   15723737 nanos:  157237.37 nanos/it
BM_stdvector_Assign:                   100 iterations in    2564985 nanos:   25649.85 nanos/it
BM_stdvector_AssignIndices:            100 iterations in    5174923 nanos:   51749.23 nanos/it
BM_stdvector_Resize:                   100 iterations in    2520054 nanos:   25200.54 nanos/it

### DIFF
--- a/Recast/Include/RecastAlloc.h
+++ b/Recast/Include/RecastAlloc.h
@@ -262,7 +262,7 @@ rcVectorBase<T, H>& rcVectorBase<T, H>::operator=(const rcVectorBase<T, H>& othe
 }
 
 template <typename T>
-class rcVector : public	rcVectorBase<T, RC_ALLOC_TEMP> {
+class rcVector : public rcVectorBase<T, RC_ALLOC_TEMP> {
 	typedef rcVectorBase<T, RC_ALLOC_TEMP> Base;
 public:
 	rcVector() : Base() {}
@@ -272,7 +272,7 @@ public:
 	rcVector(const T* begin, const T* end) : Base(begin, end) {}
 };
 template <typename T>
-class rcPermVector : public	rcVectorBase<T, RC_ALLOC_PERM> {
+class rcPermVector : public rcVectorBase<T, RC_ALLOC_PERM> {
 	typedef rcVectorBase<T, RC_ALLOC_PERM> Base;
 public:
 	rcPermVector() : Base() {}
--- a/Recast/Include/RecastAlloc.h
+++ b/Recast/Include/RecastAlloc.h
@@ -22,8 +22,6 @@
 #include <stddef.h>
 #include <stdint.h>
 
-#include "RecastAssert.h"
-
 /// Provides hint values to the memory allocator on how long the
 /// memory is expected to be used.
 enum rcAllocHint
@@ -165,7 +163,6 @@ void rcVectorBase<T, H>::reserve(rcSizeType count) {
 }
 template <typename T, rcAllocHint H>
 T* rcVectorBase<T, H>::allocate_and_copy(rcSizeType size) {
-	rcAssert(size > m_size);
 	T* new_data = static_cast<T*>(rcAlloc(sizeof(T) * size, H));
 	copy_range(new_data, m_data, m_data + m_size);
 	return new_data;
--- a/Recast/Include/RecastAlloc.h
+++ b/Recast/Include/RecastAlloc.h
@@ -20,6 +20,9 @@
 #define RECASTALLOC_H
 
 #include <stddef.h>
+#include <stdint.h>
+
+#include "RecastAssert.h"
 
 /// Provides hint values to the memory allocator on how long the
 /// memory is expected to be used.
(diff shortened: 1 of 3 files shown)


mbabinski-at-google:
I think it would be helpful to have a style guide somewhere in the documentation. C++ has had destructors for 39 years (since before it was called "C++"); I don't think eliding support for them clearly follows from avoiding "modern" C++. 

FWIW: 
    * I didn't bother with support for moves since the test suite runs on VC2010, which does not support them anyway.
    * The bulk of the complexity comes from supporting v.push_back(v[0]), while maintaining good performance, and the distinction between resize() and reserve(). IMO, both are worth it.
    * The support for destruction is only a few lines (destroy_range and call sites).
    * Likewise, supporting copying correctly adds a couple of lines at best.
    * Even if no checked-in code uses rcVector<NonPodType>, sprinkling printfs and asserts in constructors and destructors is valuable for debugging.

IMO, imitating std::vector makes sense if only because that it is the interface most likely to be familiar to a potential contributor, and decreases friction for finding bugs and sending patches.

Getting into some specifics about the interface:
    * I think `reserve` is critical to clean, fast code.
    * The least useful methods here (given that Recast doesn't support C++11) are `operator=`, `at`, `begin/end`, `front`, and `assign`.
    * I think `SmallVector` is very useful, but there are much bigger gains to be made with algorithmic changes than with small-vector optimization.

If you'd like me to include a few changes to calling code (a la #316), I'd be happy to do that.

jakobbotsch:
I took a closer look and you are right. My main concern was adding a lot of code that we won't use, but I see now after you mention it that most of the complexity is from making it robust, and that is probably fine. The alternative would be some asserts, but I think it is fine as it is.

I think we should just get rid of `at`. I also think we should delete `operator=`; if we need to copy a vector, it should be more explicit.
I think leaving begin/end is fine, as (some parts of) C++11 should eventually be allowed. I don't think people would have objections to this.

I will leave some other comments too in a review.

> If you'd like me to include a few changes to calling code (a la #316), I'd be happy to do that.

It would be nice, but it can wait to future PRs. It was mainly if we wanted to develop it for our exact use cases more incrementally.

Recast/Include/RecastAlloc.h
```diff
@@ -58,64 +59,252 @@ void* rcAlloc(size_t size, rcAllocHint hint);
...
+	void reserve(rcSizeType size);
+
+	void assign(rcSizeType count, const T& value) { clear(); resize(count, value); }
+	void assign(const T* begin, const T* end);
+
+	void resize(rcSizeType size) { resize_impl(size, NULL); }
+	void resize(rcSizeType size, const T& value) { resize_impl(size, &value); }
+
+	void push_back(const T& value);
+	void pop_back() { back().~T(); m_size--; }
```
jakobbotsch: Can we assert `m_size > 0`?

Recast/Include/RecastAlloc.h
```diff
@@ -58,64 +59,252 @@ void* rcAlloc(size_t size, rcAllocHint hint);
...
+
+	const T& operator[](rcSizeType i) const { return m_data[i]; }
+	T& operator[](rcSizeType i) { return m_data[i]; }
+	const T& at(rcSizeType i) const { return m_data[i]; }
+	T& at(rcSizeType i) { return m_data[i]; }
+
+	const T& front() const { return m_data[0]; }
+	T& front() { return m_data[0]; }
+	const T& back() const { return m_data[m_size - 1]; };
+	T& back() { return m_data[m_size - 1]; };
```
jakobbotsch: Asserts in these 8 methods above would be nice too.

Recast/Include/RecastAlloc.h
```diff
@@ -58,64 +59,252 @@ void* rcAlloc(size_t size, rcAllocHint hint);
...
+	}
+	T* new_data = allocate_and_copy(count);
+	destroy_range(0, m_size);
+	rcFree(m_data);
+	m_data = new_data;
+	m_cap = count;
+}
+template <typename T, rcAllocHint H>
+T* rcVectorBase<T, H>::allocate_and_copy(rcSizeType size) {
+	T* new_data = static_cast<T*>(rcAlloc(sizeof(T) * size, H));
```
jakobbotsch: Assert `INTPTR_MAX / sizeof(T) >= size`?
We should perhaps introduce a `calloc` variant of `rcAlloc` some time.

Recast/Include/RecastAlloc.h
```diff
@@ -58,64 +59,252 @@ void* rcAlloc(size_t size, rcAllocHint hint);
...
+	// and by ~2-5% on BM_rcVector_Push.
+	if (rcLikely(m_size < m_cap)) {
+		construct(m_data + m_size++, value);
+		return;
 	}
 
-	/// Push the specified integer onto the end of the array and increases the size by one.
-	///  @param[in]		item	The new value.
-	void push(int item) { resize(m_size+1); m_data[m_size-1] = item; }
+	rcSizeType new_cap = m_size ? 2*m_size : 1;
```
jakobbotsch: Assert `INTPTR_MAX / 2 >= m_size`

Recast/Include/RecastAlloc.h
```diff
@@ -58,64 +59,252 @@ void* rcAlloc(size_t size, rcAllocHint hint);
...
+	}
+	return *this;
+}
 
-	/// The value at the specified array index.
-	/// @warning Does not provide overflow protection.
-	///  @param[in]		i	The index of the value.
-	const int& operator[](int i) const { return m_data[i]; }
+template <typename T>
+class rcVector : public rcVectorBase<T, RC_ALLOC_TEMP> {
```
jakobbotsch: Perhaps we should call this `rcTempVector` to make it more clear the situations for which it should be used?

mbabinski-at-google: I figured "rcVector" is the common case in the code, and the few places where rcPermVector get used would get called out. Happy to change, though, if you think it would be confusing. 

Incidentally: Are you aware of users actually using different strategies for RC_ALLOC_*? It seems like it would be hard to do robustly without telling rcFree what type it is freeing.

jakobbotsch: My main problem with it is that it is unclear that you should not use such a vector for long term storage. For example, when stored as a class member you should probably use `rcPermVector`.

I am not aware of users using different strategies, but a simple and practical one would be to use arena allocation for temporary memory, and regular heap memory for permanent memory. On free, it is easy to identify whether the memory is part of the arena or not (simple pointer compares). Then you just throw the entire arena away once your meshing process is completed.

Recast/Include/RecastAlloc.h
```diff
@@ -58,64 +61,247 @@ void* rcAlloc(size_t size, rcAllocHint hint);
...
+	// and by ~2-5% on BM_rcVector_Push.
+	if (rcLikely(m_size < m_cap)) {
+		construct(m_data + m_size++, value);
+		return;
 	}
 
-	/// Push the specified integer onto the end of the array and increases the size by one.
-	///  @param[in]		item	The new value.
-	void push(int item) { resize(m_size+1); m_data[m_size-1] = item; }
+	rcAssert(RC_SIZE_MAX / (2*sizeof(T)) >= m_size);
```
jakobbotsch: I think we can leave the sizeof assertion to `allocate_and_copy`. This should just assert that we don't overflow `m_size`, i.e. `RC_SIZE_MAX / 2 >= m_size` (mostly just so the assert matches the following multiplication)

Recast/Include/RecastAlloc.h
mbabinski-at-google: Sure, but if you don't know how big the arena is up front, you will likely need to grow it on demand, presumably with allocations from the heap, and then to find out what kind of pointer you have, you need to walk a table or something.

mbabinski-at-google: Though, I guess you can just allocate an extra preamble with every rcAlloc, and use it to store info like this.

memononen: It's been a while since the allocation hints were introduced. I remember vaguely something arena based as @jakobbotsch suggests. Most integrations I've done simply ignored the hint.

Usually it has been more important to differentiate between the memory used by Recast (memory used during building) and Detour (tile data from navmesh builder).

I'm not sure if it makes sense to have perm vector class. Perm memory size is known in advance, and you should not need vector for those (maybe there's case for verts/tris, not sure).

mbabinski-at-google: There's several places where PERM memory is free'd and reallocated (see mergeContours) -- converting members to rcPermVector might help reduce some memory churn there. Additionally, using an rcPermVector instead of a naked pointer would give bounds-checking in debug builds for free.

memononen: Here's relevant discussion from google code times:
https://code.google.com/archive/p/recastnavigation/issues/78
https://code.google.com/archive/p/recastnavigation/issues/110
https://code.google.com/archive/p/recastnavigation/issues/199

It looks like I'm happy panda. The commit history should have been imported to github, maybe there is a nugget or two in the commit messages from that time.

jakobbotsch: @memononen I could see #307 solved by making `m_update` an `rcPermVector`. The alternative is to use a local `rcTempVector` in `dtTileCache::update`, but then we get allocation every time there is an update (except if we use small vector optimization). Perhaps the tiny dynamic allocation is ok since it would only happen when there were actual updates, however.

Recast/Include/RecastAlloc.h
```diff
@@ -58,64 +61,247 @@ void* rcAlloc(size_t size, rcAllocHint hint);
...
+	T* new_data = allocate_and_copy(count);
+	destroy_range(0, m_size);
+	rcFree(m_data);
+	m_data = new_data;
+	m_cap = count;
+}
+template <typename T, rcAllocHint H>
+T* rcVectorBase<T, H>::allocate_and_copy(rcSizeType size) {
+	rcAssert(RC_SIZE_MAX / sizeof(T) >= size);
+	T* new_data = static_cast<T*>(rcAlloc(sizeof(T) * size, H));
```
jakobbotsch: On second thought the assertion here actually loses us half the possible memory size, since `rcAlloc` takes `size_t`. I wonder if we should not just use `size_t` in this class and do a pass to clean up the indices.

mbabinski-at-google: I like doing signed arithmetic when possible. There's less chance of errors like "while (i-- >= 0) { ... }" (though, in fairness, modern compilers warn about that one), and while it is true that we lose half of the address space, I cannot imagine any practical case where Recast would eat >50% of it in one allocation. 

Another option is to use signed math for m_cap and m_size, but make the multiplication above unsigned. That would allow using the whole address space for a vector of anything bigger than a char.

PR #321, opened 2018-05-13
Call proper memory deallocation function in `dtDecompressTileCacheLayer`.

zapolnov:
In function `dtDecompressTileCacheLayer` buffer is allocated using a call to `dtTileCacheAlloc::alloc` (`DetourTileCacheBuilder.cpp`. line 2181). But if call to `comp->decompress` below fails, `dtFree` is used instead to deallocate the memory (line 2199, which this pull request fixes).

This happens to work for most of the users, because default implementation of `dtTileCacheAlloc::free` calls `dtFree` itself. But it breaks in rare circumstances when you use custom memory allocation *and* decompression fails.

This request fixes that.


jakobbotsch:
Thank you!

PR #322, opened 2018-05-15
Switch expandRegions to using a rcTempVector of structs rather than a blob of ints.

mbabinski-at-google:
Also updates rcVector::clear() to not require the contained type to be default-constructible.


jakobbotsch:
Nice! Thanks.

PR #323, opened 2018-05-16
Convert lvlStacks to be vectors of structs instead of a bag of ints.

mbabinski-at-google:


Recast/Source/RecastRegion.cpp
```diff
@@ -27,6 +27,14 @@
 #include "RecastAssert.h"
 #include <new>
 
+namespace {
+struct LevelStackEntry {
```
jakobbotsch: Please put brackets on a new line.

mbabinski-at-google: Done.

jakobbotsch:
Thanks!

PR #324, opened 2018-05-16
Add rcNew/rcDelete, template functions that allocate/deallocate objects using rcAlloc/rcFree.

mbabinski-at-google:
The intention is to use these to move construction logic out of rcAllocateFooObject/rcFreeFooObject and into constructors/destructors, giving users flexibility in how objects are allocated. In particular, this allows stack allocation.

Turns rcAllocateCompactHeightfield into a wrapper around rcNew as a demo. If accepted, PR's for other user objects to follow.

### DIFF
--- a/Recast/Source/Recast.cpp
+++ b/Recast/Source/Recast.cpp
@@ -149,6 +149,8 @@ rcCompactHeightfield::rcCompactHeightfield()
 	borderSize(0),
 	maxDistance(0),
 	maxRegions(0),
+	bmin(),
+	bmax(),
 	cs(0),
 	ch(0),
 	cells(0),
@@ -156,8 +158,6 @@ rcCompactHeightfield::rcCompactHeightfield()
 	dist(0),
 	areas(0)
 {
-	memset(bmin, 0, sizeof(bmin));
-	memset(bmax, 0, sizeof(bmax));
 }
 rcCompactHeightfield::~rcCompactHeightfield()
 {
--- a/Recast/Source/Recast.cpp
+++ b/Recast/Source/Recast.cpp
@@ -23,10 +23,33 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <stdarg.h>
-#include <new>
 #include "Recast.h"
 #include "RecastAlloc.h"
 #include "RecastAssert.h"
+
+namespace
+{
+/// Allocates and constructs an object of the given type, returning a pointer.
+/// TODO: Support constructor args.
+/// @param[in]		hint	Hint to the allocator.
+template <typename T>
+T* rcNew(rcAllocHint hint) {
+	T* ptr = (T*)rcAlloc(sizeof(T), hint);
+	::new(rcNewTag(), (void*)ptr) T();
+	return ptr;
+}
+
+/// Destroys and frees an object allocated with rcNew.
+/// @param[in]     ptr    The object pointer to delete.
+template <typename T>
+void rcDelete(T* ptr) {
+	if (ptr) {
+		ptr->~T();
+		rcFree((void*)ptr);
+	}
+}
+}  // namespace
+
 
 float rcSqrt(float x)
 {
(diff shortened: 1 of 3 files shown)


jakobbotsch:
I like this -- we already did this for `rcHeightfield` in #196. In fact looks like we had a dependency on `new` there, so we could've used that instead of our own variant. You can remove that include for now.

I would have no issue converting more of our structures to use constructors and follow your pattern. The only thing that irks me slightly is the verbose initialization of each member, but allowing them to be stack allocated is a worthwhile tradeoff.
@memononen what do you think of this pattern? IIRC the classes used to be stack allocatable, and then you moved to the current pattern?

memononen:
@jakobbotsch The reason I used specific functions for alloc/free was that there were some stupid cases when new would not be the right new. I think the functions are more in line with the general API style (user facing API is mostly bunch of functions, no templates used). There are not that many constructors after all. But in the end, it's cosmetic, like arguing if a singleton is global variable or not (yes it is! ;)).

```C
// TODO: Use something like std::array to store bmin/bmax?
```
Nope :) float[3] is not that such a horrible thing. If it feels nasty, `rcVec3` will give you more legibility across the project.

memononen:
@jakobbotsch Stack allocated structs not being destroyed is a good point.

If you allow/encourage folks to place `rcHeightfield` into stack, then you have to support (regular) `new` too to match the users expectations. That gets ugly with custom allocators (there might be solution, but I could not find it back then). I think that `rcNew<>` is clever way to do it, but odd and unexpected. The alloc function are not C++ like, but common enough paradigm for folks to quickly get it.

The only reason I can see to use the stack version is to automatically clean things up upon error. Most implementations I've seen eventually end up adding debug draw for the build process (and they should!), and that means that you have a "collection" struct/class which holds the different states/data of the build pipeline so that you can draw it, and that in turn means that you have some kind of destructor which handles freeing the memory. I think it is very rare that allocating the build data on stack would be a real benefit.

I think at the end of the day, it is matter of style. Alloc/free function pair is well understood pattern, as well as the new/delete & thus ctor/dtor. I think one way is enough, because the build pipeline is not much code and you don't mess with it that much once it's written. Just pick one style and stick to it :)

memononen:
@jakobbotsch You're correct about the new/delete, and it makes total sense. I'm not sure anymore why I thought that new/delete on those structs should use the Recast allocator :)

mbabinski-at-google:
My only objection to float[3] is that it is cumbersome to initialize in constructors (though this isn't an issue in c++11). rcVec3 or somesuch is a better solution anyway. Dropped TODO.

I see the main benefit of stack-allocating the user-facing structs as automatic cleanup and less boilerplate (`rcFoo* foo = rcAllocateFoo(); ... rcFreeFoo();` vs `rcFoo foo`) I don't think it is necessary to expose rcNew to users -- it is mostly just a convenience for implementing rcAlloc*(); putting it in a private header or Recast.cpp would work, too.

I hadn't noticed that <new> already gets included from a couple .cpp files. I'm totally fine using that, but note that as it stands, the includes are internal to recast, and do not affect users.

jakobbotsch:
@mbabinski-at-google If you move `rcNew` and `rcDelete` to the cpp that is fine with me.
> I'm totally fine using that, but note that as it stands, the includes are internal to recast, and do not affect users.
Right, but we still take a dependency on the STL there. Though we take many dependencies, so not really sure it matters.

mbabinski-at-google:
@jakobbotsch Sure, but there might be users who don't care what recast does internally, so long as including Recast.h doesn't conflict with their home-brewed operator ::new or whatnot.

Recast/Source/Recast.cpp
```diff
@@ -104,26 +127,44 @@ rcHeightfield::~rcHeightfield()
...
+	maxRegions(0),
+	cs(0),
+	ch(0),
+	cells(0),
+	spans(0),
+	dist(0),
+	areas(0)
+{
+	memset(bmin, 0, sizeof(bmin));
+	memset(bmax, 0, sizeof(bmax));
```
jakobbotsch: We can just value initialize `bmin` and `bmax` in the ctor:
`bmin(), bmax()`

jakobbotsch:
Thanks!

PR #326, opened 2018-05-17
Remove placement-new in RecastRegion.cpp in favor of rcTempVector.

mbabinski-at-google:


Recast/Source/RecastRegion.cpp
```diff
@@ -792,16 +791,12 @@ static bool mergeAndFilterRegions(rcContext* ctx, int minRegionArea, int mergeRe
...
 	
 	const int nreg = maxRegionId+1;
-	rcRegion* regions = (rcRegion*)rcAlloc(sizeof(rcRegion)*nreg, RC_ALLOC_TEMP);
-	if (!regions)
-	{
-		ctx->log(RC_LOG_ERROR, "mergeAndFilterRegions: Out of memory 'regions' (%d).", nreg);
-		return false;
-	}
+	rcTempVector<rcRegion> regions;
+	regions.reserve(nreg);
```
jakobbotsch: While I know that Recast is not consistently allocation-failure-safe, can we keep the checks for the large block allocations? This one might not be too bad, but for the future.
I suggest changing `reserve` to return whether it succeeded or not, and use it to retain the clean failure.

jakobbotsch:
Thanks!

PR #336, opened 2018-06-15
Attempts to make build work on Windows

jakobbotsch:
Copy of #335, Travis seemed to be stuck on an old ref...

Update Premake in Travis and AppVeyor to alpha12 and fix premake5.lua. Also ensure we treat warnings as errors.
This ought to fail because of #318.


jakobbotsch:
This is the wrong branch, oops...

Issue #334, opened 2018-06-15
build error linux


Criaan:
Hy this is Riaan try to build recast demo. Getting this error line 175 of recastalloc.h.
Comparison of integer expression of different sighnedness.

PR #337, opened 2018-06-15
Add Clang and VS2017 to CI and fix warnings

jakobbotsch:
Adds Clang and VS2017 to CI, and ensures that they are treating warnings as errors. Also fixes the new warnings that cause build break, including #334.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,3 +6,7 @@ root = true
 [*]
 indent_size = 4
 indent_style = tab
+
+[*.yml]
+indent_size = 2
+indent_style = space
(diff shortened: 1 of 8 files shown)


jakobbotsch:
It appears that the reason CI did not catch #334 in #318 is because the GCC used by Travis is so old that the warning is not reported. Clang finds the warnings successfully, so I added clang (and fixed a test failure due to floating point compares). Clang also found some unused code that I removed.

cc @grahamboree @hymerman (if you merge please rebase this instead of squashing)

PR #338, opened 2018-06-16
Fix for typo in README

pboyer:
We shouldn't leave such a shocking typo on such a nice description, now should we? ;)

This PR doesn't constitute an endorsement for passive voice, though.


jakobbotsch:
Thank you! This has been bothering me for a while.

Issue #306, opened 2018-01-12
RecastDemo: Agents spawned into the wall


hsdxpro:
Just a minor issue I noticed in the demo:
![image](https://user-images.githubusercontent.com/5315603/34872215-6718215c-f790-11e7-84c2-e3af374360b9.png)

PR #350, opened 2018-09-03
Fix agents spawned into the wall(#306)

zenoengine:
:)

### DIFF
--- a/DetourCrowd/Source/DetourCrowd.cpp
+++ b/DetourCrowd/Source/DetourCrowd.cpp
@@ -538,6 +538,11 @@ int dtCrowd::addAgent(const float* pos, const dtCrowdAgentParams* params)
 		dtVcopy(nearest, pos);
 		ref = 0;
 	}
+
+	if (!ref)
+	{
+		return -1;
+	}
 	
 	ag->corridor.reset(ref, nearest);
 	ag->boundary.reset();


jakobbotsch:
I don't think this is quite the right fix -- the function is written to handle the situation where we fail to find a poly, see eg.
https://github.com/recastnavigation/recastnavigation/blob/9f9efe943ed54a31b02270fb13b301754e918528/DetourCrowd/Source/DetourCrowd.cpp#L557-L560

If we decide that any agent should initially be placed on a valid poly (this makes absolute sense to me), then the rest of the function should also be adjusted appropriately. However an agent can also get into the `DT_CROWDAGENT_STATE_INVALID` state in other situations:
https://github.com/recastnavigation/recastnavigation/blob/9f9efe943ed54a31b02270fb13b301754e918528/DetourCrowd/Source/DetourCrowd.cpp#L972-L980

I'm not sure what the intention was here -- @memononen, how should folks handle `DT_CROWDAGENT_STATE_INVALID`? As far as I can see, when an agent gets into this state, it will need some help to recover from the user, but I'm not sure if `dtCrowd` provides a good way to teleport agents back on mesh. So (correct me if I'm wrong) pretty much the only way to recover is to remove those agents.

memononen:
It really depends on the game/simulation how to handle that kind of cases. Invalid means that the agent is too far from navigable are (can happen quite frequently in dynamic worlds).

One option is to use the navmesh query and use larger search are to find a valid polygon (with exponential retry timer!), or respawn the agent to safe location, or remove it. There's no one right solution, and the AI should be designed to deal such nasty corner cases too (been there done that, etc :)).

jakobbotsch:
But what is the mechanism for doing those things? For instance, once the agent is in an invalid state and the user has found a new poly by using larger search area, how does the user actually teleport the agent to the new position? Reach into the internals of `dtCrowdAgent` (i.e.. `corridor`)? Remove and readd the agent at the new position?

Basically my thinking is this: if people need to remove and readd an agent that is in invalid state to actually recover, then letting `addAgent` succeed when it fails to place the agent is a bit strange, since the user has to remove it again to fix the problem.

memononen:
@jakobbotsch you're correct, there should be function to set agent position (teleport). 

I never quite completed the crowd code. I had planned to take a pass on the code to make it easier to interface and sync with AI behaviors and animation, but never got there. I.e. it should be possible for the user to do his/hers own "physics" (for things like animation driven motion), or to better handle switching between navigation and non-navigation behaviors, handling off-mesh connections, etc.

PR #352, opened 2018-10-11
Island system

scripter01:
Island system: use in quick check passability from one polygon to another polygon (This system works on compares polygon island indices). 

To activate this system you need uncomment define NAVMESH_ISLAND_SYSTEM in file: "DetourNavMesh.h".
To split a mesh into islands you must use a class dtIslandManager in the file: "DetourNavMesh.h".

### DIFF
--- a/Detour/Include/DetourNavMesh.h
+++ b/Detour/Include/DetourNavMesh.h
@@ -22,6 +22,9 @@
 #include "DetourAlloc.h"
 #include "DetourStatus.h"
 
+// Uncomment to enable navmesh island system
+//#define NAVMESH_ISLAND_SYSTEM
+
 // Undefine (or define in a build cofnig) the following line to use 64bit polyref.
 // Generally not needed, useful for very large worlds.
 // Note: tiles build using 32bit refs are not compatible with 64bit refs!
@@ -149,6 +152,12 @@ enum dtPolyTypes
 /// @ingroup detour
 struct dtPoly
 {
+	dtPoly()
+	{
+#ifdef NAVMESH_ISLAND_SYSTEM
+		islandIdx = 0;
+#endif
+	}
 	/// Index to first link in linked list. (Or #DT_NULL_LINK if there is no link.)
 	unsigned int firstLink;
 
@@ -161,6 +170,20 @@ struct dtPoly
 
 	/// The user defined polygon flags.
 	unsigned short flags;
+
+#ifdef NAVMESH_ISLAND_SYSTEM
+	/// The island index.
+	unsigned int islandIdx;
+
+	/// Gets island index
+	/// @return The island index.
+	inline unsigned int getIslandIdx() const { return islandIdx; }
+
+	/// Set island index
+	///  @param[in]	idx		The new island index.
+	inline void setIslandIdx(unsigned int idx) { islandIdx = idx; }
+
+#endif
 
 	/// The number of vertices in the polygon.
 	unsigned char vertCount;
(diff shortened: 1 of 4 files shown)


Detour/Include/DetourNavMesh.h
```diff
@@ -162,6 +171,20 @@ struct dtPoly
...
+
+	/// Gets island index
+	/// @return The island index.
+	inline unsigned int getIslandIdx() const { return islandIdx; }
+
+	/// Set island index
+	///  @param[in]	idx		The new island index.
+	inline void setIslandIdx(unsigned int idx) { islandIdx = idx; }
+
+#endif
```
jakobbotsch: Instead of modifying `dtPoly`, we can simply store data per poly by using an array. You can check the nav mesh purge tool on how to do that:
https://github.com/recastnavigation/recastnavigation/blob/9f9efe943ed54a31b02270fb13b301754e918528/RecastDemo/Source/NavMeshPruneTool.cpp#L40-L131

The benefit of this is that folks do not pay for what they don't use, and you can get rid of `NAVMESH_ISLAND_SYSTEM` as a bonus.

Detour/Include/DetourNavMesh.h
```diff
@@ -635,7 +683,7 @@ class dtNavMesh
 							dtPolyRef* polys, const int maxPolys) const;
 	/// Find nearest polygon within a tile.
 	dtPolyRef findNearestPolyInTile(const dtMeshTile* tile, const float* center,
-									const float* halfExtents, float* nearestPt) const;
+									const float* extents, float* nearestPt) const;
```
jakobbotsch: Why is this change being reverted?

Detour/Source/DetourNavMesh.cpp
```diff
@@ -470,12 +470,12 @@ void dtNavMesh::connectExtOffMeshLinks(dtMeshTile* tile, dtMeshTile* target, int
 		if (targetPoly->firstLink == DT_NULL_LINK)
 			continue;
 		
-		const float halfExtents[3] = { targetCon->rad, target->header->walkableClimb, targetCon->rad };
+		const float ext[3] = { targetCon->rad, target->header->walkableClimb, targetCon->rad };
```
jakobbotsch: Reverting more changes here. I will stop commenting on these.

Detour/Source/DetourNavMesh.cpp
```diff
@@ -1302,6 +1302,9 @@ struct dtPolyState
 {
 	unsigned short flags;						// Flags (see dtPolyFlags).
 	unsigned char area;							// Area ID of the polygon.
+#ifdef NAVMESH_ISLAND_SYSTEM
+	unsigned int islandIdx;                     // Island index of the polygon
```
jakobbotsch: When changing to just use an array for this data we should simply allow the island to be serialized separately.

Detour/Source/DetourNavMesh.cpp
```diff
@@ -1520,3 +1529,51 @@ dtStatus dtNavMesh::getPolyArea(dtPolyRef ref, unsigned char* resultArea) const
...
+	if (dtStatusFailed(getPolyIslandIdx(startRef, &islandIndexFrom)))
+		return false;
+	if (dtStatusFailed(getPolyIslandIdx(endRef, &islandIndexTo)))
+		return false;
+	if (islandIndexFrom == 0 || islandIndexTo == 0)
+		return false;
+	if (islandIndexFrom != islandIndexTo)
+		return false;
+	return true;
+}
```
jakobbotsch: I have not checked your implementation closely, but I don't see how this can be correct. This will give symmetric answers, i.e. `checkPathExists(a, b) == checkPathExists(b, a)`. But that is certainly not correct in the fact of unidirectional offmesh connections.

Detour/Include/DetourNavMesh.h
scripter01: fix

Detour/Source/DetourNavMesh.cpp
scripter01: fix

Detour/Include/DetourNavMesh.h
scripter01: done

Detour/Source/DetourNavMesh.cpp
scripter01: done

Detour/Source/DetourNavMesh.cpp
scripter01: My system not work with offmesh connections.

scripter01: I think I will do this system with offmesh connections, but later/

PR #353, opened 2018-10-24
allow third-party code to define its own assert definitions

illwieckz:
The idea is to allow third-party code like Dæmon engine to define `dtAssert` and `rcAssert` as something custom (like `DAEMON_ASSERT`) and to patch Dæmon to customize those definitions on its side. If it works and if it's OK, I'll try to upstream the Detour/Recast patch so we can drop our custom _recastnavigation_ fork.


illwieckz:
Hmm, sorry, GitHub picked the wrong branch, well, one day I'll do a PR.

PR #356, opened 2018-10-31
Sync version in CMakeLists.txt with last release

mimi1vx:


jakobbotsch:
Can we make it 1.5.2 as there has been changes since 1.5.1?

jakobbotsch:
Also cc @elsid, not exactly sure how this affects people who are using this build already.

elsid:
This change shouldn't make any problems. It only changes name of dynamic library file. Proper cmake file can handle this.

mimi1vx:
and this is problem :( so it mixups with `soname`

PR #357, opened 2018-10-31
Add dependencies between libraries to allow linker to resolve symbols when build dynamic libraries

### LINKED ISSUE #355: How to build recast libraries?
Hello, 

 how to build and install Recast as library. I dont want demo or examples. Only valid libs + devel files for other project.

Preferably without alfa build systems like premake ( pretty bad idea use ALPHA build system)

elsid:
Fixes #355


jakobbotsch:
Thanks!

PR #358, opened 2018-10-31
Allow to use recastnavigation by find_package and add_subdirectory

elsid:
* When use `add_subdirectory` it is no more required to say cmake where to find include directories. 
* Allow to use cmake namespace `RecastNavigation` to link libraries. Ex:
```cmake
target_link_libraries(myapp
    RecastNavigation::DebugUtils
    RecastNavigation::Detour
    RecastNavigation::Recast
)
```


jakobbotsch:
Thank you.

Issue #343, opened 2018-08-14
find path enters an infinite loop.


betterdenger:
Hi,

    My test input as bellow,
```
float start[3] = { -22.636648178101, 12.46023273468, -123.88012695313 };
float end[3] = { -12.93, 12.08, -110.27 };
float speed = 0.77999997138978;
```

    And the target test navmesh binary as bellow,
[all_tiles_navmesh.zip](https://github.com/recastnavigation/recastnavigation/files/2366454/all_tiles_navmesh.zip)

    If i am trying find a smooth path use these data, i will enter an infinite loop from a iteration. From smooth_path_count_ = 17, the loop keeps give me bad point (-14.4847, 12.4086, -113.422).

    Is there any hints for me?

Thanks

robinxb:
same problem

jakobbotsch:
I have been unable to get your navmesh file to load correct, it seems like it is corrupt in some way. When loading it fails to load any tiles because all tile-refs are 0.

MINIONBOTS:
I can also confirm I had this several times, very rare though. It happens for me when I set the additional HScale value of the pathfind algorightm too high.  I was unable to completely fix it on my end, added a hacky failsafe, but afaik it happens because the node weights are using not only the 'already traveled distance' as costs but in addition such a HScale value. This kinda made the pathfinding rotate between a few nodes until you run out of available nodes...

jackpoz:
I have the same issue, with detour entering an infinite loop in getPathToNode() looping through nodes with circular reference (A -> B -> C -> A) https://github.com/recastnavigation/recastnavigation/blob/master/Detour/Source/DetourNavMeshQuery.cpp#L1212
>dtNavMeshQuery::getPathToNode (this=0x7ffefeeebce0, endNode=0x7ffefe5688a0, path=0x7ffefdbe1810, pathCount=0x7fff029fae10, maxPath=71) at /dep/recastnavigation/Detour/Source/DetourNavMeshQuery.cpp:1212
dtNavMeshQuery::findPath (this=0x7ffefeeebce0, startRef=13510831094374714, endRef=13510831094374726, startPos=0x7fff029fae14, endPos=0x7fff029fb3f0, filter=0x7ffefdbe1ac0, path=0x7ffefdbe1810, pathCount=0x7fff029fae10, maxPath=71) at /dep/recastnavigation/Detour/Source/DetourNavMeshQuery.cpp:1193

[...]

PR #364, opened 2018-11-08
optimize dtClosestHeightPointTriangle

cloudwu:
dot00 == v0[0] * v0[0] + v0[2] * v0[2]
...
Denom  = dot00 * dot11 - dot01 * dot01
= (v0[0] * v0[0] + v0[2] * v0[2]) * (v1[0] * v1[0] + v1[2] * v1[2]) - (v0[0] * v1[0] + v0[2] * v1[2]) * (v0[0] * v1[0] + v0[2] * v1[2])
= (v0[0] * v1[2] - v0[2] * v1[0]) * (v0[0] * v1[2] - v1[0] * v0[2])
...

Both u and v have (v0[0] * v1[2] - v1[0] * v0[2]), so we can simplify them.

This optimization can also keep more precision by reduce multiplication.

PR #373, opened 2019-02-08
Add guard check against infinite loop in detour.

jackpoz:
Add a guard check against infinite loop occurring in detour, happening when there is a circular reference in m_nodePool, i.e. A -> B -> C -> A.

While this might be caused by another issue in the code, it's better to just fail the path instead of entering an infinite loop.

As every node has only 1 reference to another node, the max number of iteration we can do before we are sure we are in an infinite loop is m_nodePool->getMaxNodes() . This also ensures that we don't accidentally exit too early for huge m_nodePool sizes.

Tested this code in a project where we have been experiencing this issue after upgrading from https://github.com/recastnavigation/recastnavigation/commit/2c85309280dbc9c82029e7ab16dfb01b9235c74e to https://github.com/recastnavigation/recastnavigation/commit/14b2631527c4792e95b2c78ebfa8ac4cd3413363

Ref https://github.com/recastnavigation/recastnavigation/issues/343

### DIFF
--- a/Detour/Source/DetourNavMeshQuery.cpp
+++ b/Detour/Source/DetourNavMeshQuery.cpp
@@ -1208,6 +1208,9 @@ dtStatus dtNavMeshQuery::getPathToNode(dtNode* endNode, dtPolyRef* path, int* pa
 	int length = 0;
 	do
 	{
+		// Go through the whole m_nodepool max once, otherwise it's most likely a sign of infinite loop
+		if (length > m_nodePool->getMaxNodes())
+			return DT_FAILURE;
 		length++;
 		curNode = m_nodePool->getNodeAtIdx(curNode->pidx);
 	} while (curNode);


jakobbotsch:
This is a band-aid that does not fix an underlying problem (if there is one). In fact I think it makes things worse since now we are potentially hiding issues that may be there.
Instead I would like to identify what the underlying issue that people are running into and then hopefully improve on that instead. Here are some common pitfalls that you should check:
1. Bad threading
2. Bad `dtQueryFilter` costs -- costs should not be less than 1

I have used R&D in projects where we did thousands of queries every second without running into these issues so I think there are some complex interactions necessary to run into this isssue. If we could get some kind of way to reproduce this issue it would of course make it much better.

memononen:
These type of infinite loops in the past have always been about bad weights.

I wonder if it would make sense to add `dtMin(1.0f, m_query.filter->getCost(...))` on all uses of  `getCost()`? Or maybe it's even a case for an assert? I suspect it will not affect performance, but needs testing of course.

Maybe known pitfals like these could have upper bound, like in the PR, but you would assert (with comment) instead of breaking gracefully. That way you catch problems during dev and there would be instructions how to fix it.

jakobbotsch:
Asserts would be good. I don't think `dtMax` would be a good idea, since that just hides the problem from the user again.
`dtQueryFilter` could also scale its costs automatically if the user adds costs that are less than 1, but this adds runtime overhead.

jackpoz:
We don't set the costs so they are set to the default 1.0f in dtQueryFilter().
This doesn't mean that getCost() won't return a value < 1.0f as it's defined as https://github.com/recastnavigation/recastnavigation/blob/master/Detour/Source/DetourNavMeshQuery.cpp#L99 so for very close pa and pb it becomes very small (is that what you meant ?)

jakobbotsch:
No, the important part is that the heuristic remains admissible (i.e. it should not overestimate the cost to the target). The heuristic is simply distance * H_SCALE: https://github.com/recastnavigation/recastnavigation/blob/14b2631527c4792e95b2c78ebfa8ac4cd3413363/Detour/Source/DetourNavMeshQuery.cpp#L1153

If the query filter costs are less than `H_SCALE` the cost of the shortest path can be smaller than the heuristic, but both calculations use the distance.

Since you do not modify the costs it would be very helpful if you could provide a way to reproduce this issue in the demo (or with a small example). That will help a lot in getting to the bottom of this issue.

jackpoz:
Found some strange values while debugging a full memory dump when the issue happened:
image

This suggests dtNavMeshQuery::closestPointOnPoly() somehow put a nan in the Y .
As we upgraded from https://github.com/recastnavigation/recastnavigation/commit/2c85309280dbc9c82029e7ab16dfb01b9235c74e to https://github.com/recastnavigation/recastnavigation/commit/14b2631527c4792e95b2c78ebfa8ac4cd3413363 (full list of changed code at https://github.com/TrinityCore/TrinityCore/commit/5ff88ea04aec4677f1c1d669674e5442288a25e3#diff-e9e8feb77bdeca9109535294c4c1f8a3 ), I would say the commit that changed the behavior here is https://github.com/recastnavigation/recastnavigation/commit/7ccb72b3839c7e441b0b83cfb28d54231fd6495e

jakobbotsch:
If you could obtain the vertices of that polygon and the end point then that should give all the info necessary to reproduce that.
Essentially the contents of `verts` and `pos` at this point:
https://github.com/recastnavigation/recastnavigation/blob/14b2631527c4792e95b2c78ebfa8ac4cd3413363/Detour/Source/DetourNavMeshQuery.cpp#L541

jackpoz:
I added a few isfinite() checks in that function, here's the locals at https://github.com/recastnavigation/recastnavigation/blob/14b2631527c4792e95b2c78ebfa8ac4cd3413363/Detour/Source/DetourNavMeshQuery.cpp#L583

>closest[0] : 642.133301
closest[1] : 609.980835
closest[2] : 5823.99951

>v : 0x000000cfed8fcd68 {0x000002b931c974a4 {642.133301}, 0x000002b931c97498 {641.066589}, 0x000002b931c9751c {...}}
    [0x00000000]: 0x000002b931c974a4 {642.133301}
    [0x00000001]: 0x000002b931c97498 {641.066589}
    [0x00000002]: 0x000002b931c9751c {641.866638}
	
>h : -nan(ind)

Running again dtClosestHeightPointTriangle() dtClosestHeightPointTriangle with the input values, this is what we get:
>v0 : {-0.266662598, 0.000000000, 0.000000000}
v1 : {-1.06671143, 0.000000000, 0.000000000}
v2 : 0x000000cfed8fcb68 {0.000000000, 0.000000000, 0.000000000}

>denom : 0.000000000
u : 0.000000000
v : -0.000000000

>epsilon : -0.000000000

Condition https://github.com/recastnavigation/recastnavigation/blob/14b2631527c4792e95b2c78ebfa8ac4cd3413363/Detour/Source/DetourCommon.cpp#L228 is true as 0.0 > -0.0 is true so at https://github.com/recastnavigation/recastnavigation/blob/14b2631527c4792e95b2c78ebfa8ac4cd3413363/Detour/Source/DetourCommon.cpp#L229 there is
>h = 609.980835 + (0 * 0 + 0 * -0) / 0

the "0" part gives -nan(ind) and 609 + nan is still nan

jackpoz:
I ran a test with the same input locally, the old code https://github.com/recastnavigation/recastnavigation/commit/7ccb72b3839c7e441b0b83cfb28d54231fd6495e#diff-4ee8a41982c1c5ad28279226a4950e31L229 just set h to a[1] as everything else was 0

jakobbotsch:
> 	float a[3] = { 642.133301, 609.980835, 5823.99951 };
	float b[3] = { 641.066589, 609.980835, 5823.99951 };
	float c[3] = { 641.866638, 609.980835, 5823.99951 };

This is a degenerate triangle so the underlying problem might be that Recast is generating bad data. Unfortunately this requires going even further back to see what data was passed to Recast. As a first step it would be nice with the vertices of the polygon itself (that produces this bad triangulation), i.e. the verts iterated over in the loop in `closestPointOnPoly`.

The reason it "worked" before is because floating point inaccuracies caused `invDenom` to become a very large finite value (instead of infinity). Then `u = 0 * invDenom` is 0 instead of `nan` if `invDenom` was infinity.

I think the most practical fix is to check if `denom` is close to 0 and then just linearly interpolate the height along the line segment. @memononen what do you think? And is Detour supposed to be robust to such intricacies in the data (collinear segments in the polys)?

jackpoz:
I would suggest to start reverting https://github.com/recastnavigation/recastnavigation/commit/7ccb72b3839c7e441b0b83cfb28d54231fd6495e since what was supposed to be an optimization is actually not handling a float division by zero case.
In addition to that I would add a dtAssert(std::isfinite(h)) call in dtClosestHeightPointTriangle() to ensure the issue gets spotted sooner without having to go through to what I had to go through.

After that we can investigate more into recast.

jakobbotsch:
> I would suggest to start reverting [7ccb72b](https://github.com/recastnavigation/recastnavigation/commit/7ccb72b3839c7e441b0b83cfb28d54231fd6495e) since what was supposed to be an optimization is actually not handling a float division by zero case.

Reverting this is fixing your problem but reintroducing the problem fixed in #364 (caused by floating point inaccuracy). This is not an optimization for speed -- it is an optimization for accuracy.
The previous version is not really more "correct" in this case, in that does not handle the 0 case either. You just have a degenerate triangle where the floating point inaccuracy makes it seem like it works. If you translate the triangle closer to 0 I expect you will get different results (even though it is the same triangle) on the old version.

Definitely agree on the asserts. Also, `dtNavMeshQuery::findPath` should validate that the input is a well-formed point, which would immediately have failed your path finds, even without assertions enabled. I will add these.

jackpoz:
@jakobbotsch something strange happened indeed in Recast, this is how the result look
image

I will investigate further to check how that happened and possibly set up a test case. It also makes it quite easy to spot which commit changed the behavior, just need to try a few recast commits.

Closing this PR as you are going to add anyway the checks :)
Should we continue the discussion at https://github.com/recastnavigation/recastnavigation/issues/343 or somewhere else ?

jakobbotsch:
Indeed I would expect the problem to be in Recast since Detour ends up with the degenerate triangle. But one question is whether Detour should be more robust to these scenarios.

We can continue in #343 👍

PR #374, opened 2019-02-10
Add consistent input validation to dtNavMeshQuery

jakobbotsch:
Validate input values more consistently and completely, including that points are finite.
This would have saved everyone some time in #343/#373.

### DIFF
--- a/Detour/Source/DetourNavMeshQuery.cpp
+++ b/Detour/Source/DetourNavMeshQuery.cpp
@@ -1301,8 +1301,8 @@ dtStatus dtNavMeshQuery::initSlicedFindPath(dtPolyRef startRef, dtPolyRef endRef
 	m_query.raycastLimitSqr = FLT_MAX;
 	
 	// Validate input
-	if (!m_nav->isValidPolyRef(startRef) || !m_nav->isValidPolyRef(endRef)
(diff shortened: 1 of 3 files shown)


Detour/Source/DetourNavMeshQuery.cpp
```diff
@@ -2070,13 +2101,17 @@ dtStatus dtNavMeshQuery::moveAlongSurface(dtPolyRef startRef, const float* start
...
+	if (!m_nav->isValidPolyRef(startRef) ||
+		!startPos || !dtVisfinite(startPos) ||
+		!endPos || !dtVisfinite(endPos) ||
+		!filter || !resultPos || !visited || !visitedCount ||
+		maxVisitedSize <= 0)
+	{
 		return DT_FAILURE | DT_INVALID_PARAM;
+	}
+
+	*visitedCount = 0;
```
memononen: I used to have rule that output should always be valid (i.e. setting return array counts to zero), even when returning failure. I think that still held after `dtStatus` was introduced. Might not have been consistently implemented though. Some code might rely on that.

jakobbotsch: Will make sure this is still the case then.

PR #381, opened 2019-03-03
Robust nearest point queries

jakobbotsch:
@memononen is this along the lines of what you meant?


Detour/Source/DetourNavMeshQuery.cpp
```diff
@@ -514,88 +514,14 @@ dtStatus dtNavMeshQuery::findRandomPointAroundCircle(dtPolyRef startRef, const f
 dtStatus dtNavMeshQuery::closestPointOnPoly(dtPolyRef ref, const float* pos, float* closest, bool* posOverPoly) const
```
memononen: Should update `dtNavMeshQuery::getPolyHeight` too.

Detour/Source/DetourNavMesh.cpp
```diff
@@ -639,41 +675,26 @@ void dtNavMesh::closestPointOnPoly(dtPolyRef ref, const float* pos, float* close
...
-			{
-				dmin = edged[i];
-				imin = i;
-			}
-		}
-		const float* va = &verts[imin*3];
-		const float* vb = &verts[((imin+1)%nv)*3];
-		dtVlerp(closest, va, vb, edget[imin]);
+		// Clamp to nearest detail edge.
+		closestPointOnDetailEdge(tile, poly, pos, closest);
```
memononen: This should work, but I'd test that it returns same-ish point as previous code.

Detour/Source/DetourNavMesh.cpp
```diff
@@ -616,12 +616,48 @@ void dtNavMesh::baseOffMeshLinks(dtMeshTile* tile)
...
+	const unsigned int ip = (unsigned int)(poly - tile->polys);
+	const dtPolyDetail* pd = &tile->detailMeshes[ip];
+
+	float result[3];
+	dtVcopy(result, pos);
+
+	float dmin = FLT_MAX;
+	for (int i = 0; i < pd->triCount; i++)
+	{
+		const unsigned char* tris = &tile->detailTris[(pd->triBase + i) * 4];
```
memononen: tris[3] is flag indicating if the edge is part of the nav poly contour, see:
https://github.com/recastnavigation/recastnavigation/blob/master/DebugUtils/Source/DetourDebugDraw.cpp#L104

With that information, it should be possible to iterate over edges only once, by checking that either the contour flag is set, or indices of a segment are in ascending order. Or iterate only the boundary edges.

memononen: Note: The border flag is not 100% accurate, there can be false positives, which should be ok in this case.

Detour/Source/DetourNavMesh.cpp
jakobbotsch: I wonder if this is even better since we are now unconditionally iterating all detail tri edges, while before we had an early out (in code below) if we found the containing triangle. I wonder which is better (with no epsilon it is probably likely that all the triangle checks below fail, however).

Detour/Source/DetourNavMeshQuery.cpp
jakobbotsch: Right. Not sure what the best way is -- if internal points can fail all the triangle checks then we need an additional poly check to figure out whether the point is inside the poly. At that point it is basically the same as `closestPointOnPoly`.

Detour/Source/DetourNavMesh.cpp
jakobbotsch: Gave it a go. I split them into two functions `closestPointOnDetailedContour` and `closestPointOnDetailEdge`. The latter could be used for both cases, but the common case (where point is outside poly) needs the former, and it can be written more efficiently. I will do some benchmarks and see whether it makes a difference. How does it look?

memononen: May I suggest `closestPointOnDetailEdge` -> `closestPointOnDetailEdges`, as it does iterate over all edges, not just one. Maybe `closestPointOnDetailedContour` -> `closestPointOnDetailContour` for consistency.

DebugUtils/Source/DetourDebugDraw.cpp
```diff
@@ -101,7 +101,8 @@ static void drawPolyBoundaries(duDebugDraw* dd, const dtMeshTile* tile,
 				}
 				for (int m = 0, n = 2; m < 3; n=m++)
 				{
-					if (((t[3] >> (n*2)) & 0x3) == 0) continue;	// Skip inner detail edges.
+					int edgeFlags = (t[3] >> (n * 2)) & 0x3;
```
memononen: You could make this a helper function too, as it's a little obscure, so you could just call:
```C
if ((rcDetailEdgeFlags(t[3], n) & DT_DETAIL_EDGE_CONTOUR) == 0) continue;
```

Detour/Source/DetourNavMesh.cpp
```diff
@@ -616,12 +616,102 @@ void dtNavMesh::baseOffMeshLinks(dtMeshTile* tile)
 	}
 }
 
+void dtNavMesh::closestPointOnDetailedContour(const dtMeshTile* tile, const dtPoly* poly, const float* pos, float* closest) const
```
memononen: There's already `closestPointOnPolyBoundary` so this should be `closestPointOnDetailBoundary` for consistency.

Detour/Source/DetourNavMesh.cpp
memononen: I like this behavior better, since this is common case where the raycasts fail. My point was that just make sure the `closestPointOnDetailEdge` returns similar results as previous code since it is using different data. I believe it should work, but just double check :)

Detour/Source/DetourNavMeshQuery.cpp
jakobbotsch: This code seems is a bit weird: https://github.com/recastnavigation/recastnavigation/blob/3a619d773deb7e3a15ee215217c825995fe71312/Detour/Source/DetourNavMeshQuery.cpp#L680-L689
For instance, if you call it with a point that is far away, it returns approximately the mid point. Shouldn't this just be using `dtDistancePtSegSqr2D` too?

Detour/Source/DetourNavMesh.cpp
jakobbotsch: Should the flag be `DT_DETAIL_EDGE_BOUNDARY` too?

Detour/Source/DetourNavMesh.cpp
jakobbotsch: I have added just a single closestPointOnDetailEdges taking a template parameter that controls the behavior.

In terms of speed, iterating only boundary edges makes a massive difference (around 100% faster to iterate only boundary edges vs all edges).

In comparison to the old, results are a bit mixed. I tested with 10 million closestPointOnPoly queries on a large poly (11 detail triangles) and small poly (1 triangle).
When the point is inside, the new version is about 20% faster for both polys, I guess because we no longer calculate closest "non-detailed" edge.

When the point is outside, the old version is quite a lot faster on the big poly, depending on which edge is closest, with the old version takes between 55% and 90% of the time of the new version.
For the small poly the new version is 0-20% faster depending again depending on the edge.

Anyway, since this is a bug fix I guess the trade off is worth it. I also think it is likely that this function is called with points inside the poly more often than outside.

Detour/Source/DetourNavMeshQuery.cpp
memononen: Yes, it should.

Detour/Source/DetourNavMesh.cpp
memononen: I think those numbers makes sense. Did you check how often the `dtClosestHeightPointTriangle` failed and the edge fall back was used for the inside case?

Detour/Source/DetourNavMeshQuery.cpp
jakobbotsch: On the other hand I am not sure how useful this behavior is (returning true for off-mesh connections). I have actually broken it right now. Oh well, I will restore it before merging this.

Detour/Source/DetourNavMesh.cpp
jakobbotsch: I did not try "provoking" this situation and never saw it happen, but I was clicking manually and that makes it really unlikely to happen. I can try with some edge/junction points tonight and see if they trigger the case.

memononen: I used to have a test level which had big polygons on an undulating terrain quite far from origo. It might had had some big sliver polygons too. I think that level had problems inside the polygon. I tested it using a dozen agents.

Detour/Source/DetourNavMesh.cpp
memononen: Yes. Contours are the specific outlines before triangulation.

Detour/Source/DetourNavMesh.cpp
jakobbotsch: I tried the following situation and moving the target around a bit, but still didn't trigger the case:
image

This is with origin at 5000, 0, 5000.

memononen: You could try bigger tile size (i.e. 4 times bigger) and add few obstacles in there so that you're more likely to get sliver nav polys.

jakobbotsch: I was able to hit the case after adding some obstacles. However, it is not hitting it internally in polygons, in all the cases it was on the border of a polygon (i.e. `dtPointInPolygon` disagrees with the triangle check in `dtClosestHeightPointTriangle`). That is, I could never get a breakpoint inside the if here to be hit:
```cpp
float closest[3];
closestPointOnDetailEdges<false>(tile, poly, pos, closest);
*height = closest[1];

float closest2[3];
closestPointOnDetailEdges<true>(tile, poly, pos, closest2);
if (dtVdistSqr(closest, closest2) > 0.0001 * 0.0001)
{
	int ii = 0; // never hit
}
```
I think I will leave it as is since this should be the most robust way to do it and it almost never happens anyway. I have also added the mesh I have made to test to this PR.

PR #405, opened 2019-09-10
Add vcpkg installation instructions

grdowns:
Recast is available as a port in vcpkg. Adding installation instructions will help users get started by providing a single command they can use to build Recast and include it into their projects


grahamboree:
Thanks!  We'll try to keep the vcpkg version up to date as we make new releases

PR #406, opened 2019-09-25
Updates the README instructions for linux

Dig-Doug:
The binary has to be run from the bin directory, otherwise the DroidSans font can't be found and shows the error "Could not init GUI renderer.":

https://github.com/recastnavigation/recastnavigation/blob/master/RecastDemo/Source/main.cpp#L129


votiakovbogdan:
Thank You! you' ve saved me a lot of time!

PR #409, opened 2019-10-21
Explicitly define winding as being signed

icecream95:
On some architectures (e.g. ARM), char is unsigned, so winding should be explicitly defined as signed char.

I grepped through the source code, and there doesn't seem to be anywhere else where this is needed.


jakobbotsch:
Thanks!

PR #418, opened 2020-01-29
Bounds checks were unnecessary inclusive.

Learko:
I believe using `>=` in this case is incorrect. `rcOffsetPoly` returns at most N\*2 vertices, thus you allocating N*2 vertices for result and passing `maxOutVerts = N*2`, but if each vertex will be translated into to two vertices, function will return 0, even though buffer isn't full yet.


grahamboree:
Thanks!

Issue #424, opened 2020-03-12
Method dtNavMesh::connectExtLinks might produce incorrect value for link->bmax with /fp:fast or -ffast-math optimization.


Learko:
There's a problem in calculation of `tmin/tmax` values(https://github.com/recastnavigation/recastnavigation/blob/master/Detour/Source/DetourNavMesh.cpp#L442), some compilers like MSC with `/fp:fast` or clang with `-ffast-math` optimize this:
```
float tmin = (neia[k*2+0]-va[0]) / (vb[0]-va[0]);
float tmax = (neia[k*2+1]-va[0]) / (vb[0]-va[0]);
```
into something like this:
```
float tmin = neia[k*2+0]-va[2];
float tmax = neia[k*2+1]-va[2];

float d = 1.0f / (vb[2]-va[2]);

tmin *= d;
tmax *= d
```

After that you convert `tmin/tmax` to `unsigned char`.

And the problem is when `neia[k*2+0] == vb[0]`, due to precision loss in calculating inversed denominator, `tmax` won't be always equal 1. For instance if `neia[k*2+1] == vb[0] == 160.0f` and `va[0] == 139.5f`, `tmax` will be `0.99999994f` and `link->bmax` 254 instead of 255.

It doesn't seem to be a big deal, until you try to raycast through the edge of tile, like this: 
![poc](https://user-images.githubusercontent.com/8157199/76519774-6cf96b00-6483-11ea-8d71-6d2cecd7a49c.png) (raycast from the right point to left, poly in top-right tile has `bmax == 254`).

In `dtNavMeshQuery::raycast` (https://github.com/recastnavigation/recastnavigation/blob/master/Detour/Source/DetourNavMeshQuery.cpp#L2565) you check `link->bmin == 0 && link->bmax == 255`, but `bmax` here is 254 thus later on it detects a "hit".

If you google for "recast detour "-ffast-math"", you'll find that I'm not the only one compiling this library with this flag, so it would be nice to fix it somehow. Right now I've come up with couple ways of "fixing" it:
1. You can either mark `tmin/tmax` as `volatile`, which "should" work on all platforms, but it's gonna make it much slower.
2. You can explicitly check cases `neia[k*2+0] == va[0]` and `neia[k*2+0] == vb[0]`, it seems to trick MSC compiler, but I'm not sure that it'll work everywhere.
[...]

PR #428, opened 2020-04-17
Fix rounding error due to precision loss after optimization.

Learko:
Fix for #424 .


grahamboree:
Makes sense.  Thanks for this!

PR #430, opened 2020-05-08
Iterative straight path building without the need for result arrays

niello:
The main idea is to get rid of preallocated fixed-size arrays for **findStraightPath** results and to limit straight path length to amount required by user. Desired point count may depend on the situation, including previous point flags etc. The new approach allows to get points one by one until the user decides to stop or until the end is reached.

Important improvement is that the user can now control options (DT_STRAIGHTPATH_*_CROSSINGS) on per-point basis.

Doesn't affect existing code, only extends API. Produces exactly the same results as previous implementation (tested in RecastDemo in various circumstances). Performance is also the same.

There is another branch:
https://github.com/niello/recastnavigation/tree/dev/iterative_straight_path_compare
where old and new functions work side by side and output is compared bitwise.

Usage example:
https://github.com/niello/deusexmachina/blob/master/DEM/Game/src/AI/Movement/SteerAction.cpp
https://github.com/niello/deusexmachina/blob/07c2f69471b6be6e339c7798597d0dd8dbbac922/DEM/Game/src/AI/Navigation/NavigationSystem.cpp#L289

### DIFF
--- a/Detour/Include/DetourNavMeshQuery.h
+++ b/Detour/Include/DetourNavMeshQuery.h
@@ -159,6 +159,30 @@ public:
 	virtual void process(const dtMeshTile* tile, dtPoly** polys, dtPolyRef* refs, int count) = 0;
 };
 
+/// Context required for iterative straight path building.
+/// Search args are stored inside for consistency between calls.
+/// @ingroup detour
+struct dtStraightPathContext
+{
+	const dtPolyRef* path;
+	float endPos[3];
+	int pathSize;
+
+	struct Corner
+	{
+		dtPolyRef polyRef;
+		float point[3];
+		int fromIndex;
+		unsigned char flags;
+	};
+
+	float lastCorner[3];
+	int lastCornerIndex;
+	Corner corner;
+	int i;
+	bool cornerFound;
+};
+
 /// Provides the ability to perform pathfinding related queries against
 /// a navigation mesh.
 /// @ingroup detour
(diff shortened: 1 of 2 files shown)


vorlov-playrix:
@memononen , does anyone own this repo now? It looks abandoned, there was no reaction on my PR for 5 months.

jakobbotsch:
We are a few (supposed) maintainers, but unfortunately personally I stopped using the library a while ago and have not been doing a good job with the maintenance since then. We should probably look for new maintainers again.

As for this PR: I'm a bit confused how it compares to what already exists. `findStraightPath` already returns as many points as the user asks so why not just ask for as few points as you need? This is how `dtPathCorridor` works which is optimized to be as iterative as possible because replanning may often be necessary.

vorlov-playrix:
That's because you may not know in advance how many points you want, and you don't want to request maximum possible count (or this maximum is unpredictable too).
My own example is evaluating actions from path. Let each area have traversal action associated. When the path corridor is built, my agent detects next corner or area change point. If area changed, but direction and traversal action are the same, it wants to continue. If direction changes, but action is still the same, it wants the next point for smooth turning. If action changes, it may want to brake or to do something special. And I can't know how many points agent will examine until the action is built.
The solution proposed allows you to iterate the path no further than you really require to, and saves you from using any kind of intermediate array. Performance impact is negligible as of my experience.
You may look at the code with described logic if you want, but I believe that brief details above are enough to get the idea.

jakobbotsch:
It's not that you need to know how many points you will end up needing; `dtCrowd` agents do not know this either. Rather, they just request a few points, do some optimization, and steer towards the current point. If replanning is necessary or when the agent gets closer to the corner the path corridor is readjusted and then a few more target points are retrieved with `findStraightPath`.
See here: https://github.com/recastnavigation/recastnavigation/blob/9337e124182697de93acb656ef25766486738807/DetourCrowd/Source/DetourCrowd.cpp#L1109-L1111
This is a tiny number, only 4:
https://github.com/recastnavigation/recastnavigation/blob/9337e124182697de93acb656ef25766486738807/DetourCrowd/Include/DetourCrowd.h#L34-L39

That's because `findStraightPath` is already usable in an iterative manner (and even the full path finding, not just string pulling, can be done in a more iterative manner by using `dtPathCorridor`). I would advise that you check if `dtPathCorridor` covers your use case. [There is some documentation in the source](https://github.com/recastnavigation/recastnavigation/blob/9337e124182697de93acb656ef25766486738807/DetourCrowd/Source/DetourPathCorridor.cpp#L158-L198).

vorlov-playrix:
Yes, I know really well how it works, because I disassembled all the dtCrowd agent code into pieces while designing my own navigation agent)

Initially I implemented the same thing with 3 or 4 corners, but it is not what I need:
Sometimes the first corner changes a traversal action, so I don't need the next one. So why to spend time calculating all 4?
Sometimes dense edges may produce multiple corners in a relatively short linear path segment, and by requesting only first 4 I can't detect that there is already a time to start braking or turning.

So no, requesting a fixed number of corners is not the same, it offers much less flexibility without clear pros.
If you believe me that this is really useful at least in my case, what are another reasons not to accept this PR?

jakobbotsch:
> Initially I implemented the same thing with 3 or 4 corners, but it is not what I need:
Sometimes the first corner changes a traversal action, so I don't need the next one. So why to spend time calculating all 4?
Sometimes dense edges may produce multiple corners in a relatively short linear path segment, and by requesting only first 4 I can't detect that there is already a time to start braking or turning.

You might consider looking ahead in the path corridor to check if you are approaching a polygon of the right type. You should already have the full corridor if I understand correctly (and usually this is the most expensive part of path finding anyway).

> So no, requesting a fixed number of corners is not the same, it offers much less flexibility without clear pros.
> If you believe me that this is really useful at least in my case, what are another reasons not to accept this PR?

The pros is that the code is simpler and already there.
Even if this is useful, we at least need to not regress performance as this PR does. For example, simple example on `nav_test.obj`:
```c++
auto time = GetTickCount64();
for (int i = 0; i < 1000000; i++)
{
	m_navQuery->findStraightPath(m_spos, epos, m_polys, m_npolys,
		m_straightPath, m_straightPathFlags,
		m_straightPathPolys, &m_nstraightPath, MAX_POLYS, m_straightPathOptions);
}
time = GetTickCount64() - time;
printf("%lld\n", time);
```
Before this change: 1328 ms
After this change: 1563 ms
So this change regresses performance of `findStraightPath` by 15% on my machine.

vorlov-playrix:
It is better to measure in real code that is rewritten to take advantage of not calculating unused corners. Also you should measure how the time spent is structured. I'm almost sure there is a repeated re-initialization part that incurs that difference, iteration logic is almost the same as before.

What of corridor checking, imagine I have checked it and detected that 10 polys farter ahead the action changes. How this helps me if I have corner[4] array compiled into my binary? Dynamic array is not good, so I should create significantly bigger stack array just in case. But what if there is a corner closer than that 10'th poly and therefore I don't need all next points? This can be known only after some string pulling, but I already requested many corners that are wasted.

But OK, we spend much time discussing this solution and I have no enough motivation to prove it further. You may close this PR if you see no use of it. No offence, really, just too much cost for a free contribution.

jakobbotsch:
> It is better to measure in real code that is rewritten to take advantage of not calculating unused corners. Also you should measure how the time spent is structured. I'm almost sure there is a repeated re-initialization part that incurs that difference, iteration logic is almost the same as before.

There are hundreds of existing projects using Recast & Detour that will not be updated like this, I am also worried about other projects that won't make such a switch.

> What of corridor checking, imagine I have checked it and detected that 10 polys farter ahead the action changes. How this helps me if I have corner[4] array compiled into my binary? Dynamic array is not good, so I should create significantly bigger stack array just in case. But what if there is a corner closer than that 10'th poly and therefore I don't need all subsequent points? This can be known only after some string pulling, but I already requested many corners that are wasted.

Ok, fair enough, I recognize that your use case is not covered well by only requesting few points since you might need to react early to points several turns down-the-road.

> But OK, we spend much time discussing this solution and I have no enough motivation to prove it further. You may close this PR if you see no use of it. No offence, really, just too much cost for a free contribution.

Ok; thank you anyway. I think if you want this to get farther you should try doing it without regressing `findStraightPath` for other people. Some code duplication would be ok (we already do this for `findPath` and `updateSlicedFindPath`).

niello:
Really good idea, though I can't promise I have time or reason to do this in the near future. Will keep it in mind.

PR #437, opened 2020-06-16
Several fixes for make install

DarthGandalf:
To make this easier to package


DarthGandalf:
Perhaps instead of "Don't install DebugDraw.h" it should install every header to /usr/include/recastnavigation ?

DarthGandalf:
I'll answer my question: DebugDraw.h is used by OpenMW, so can't be just deleted...

DarthGandalf:
@jakobbotsch hi?

jakobbotsch:
@DarthGandalf Thank you, sorry for the late reaction!

PR #443, opened 2020-10-05
Improve debug builds for MSVC

Bromeon:
Two changes that make life easier when using the _Visual Studio_ and _nmake_ generators in CMake:

1. When compiling both Release/Debug (e.g. as a batch build inside Visual Studio), it no longer overwrites libraries .lib files. Instead, it creates distinct ones for Release (the same as before), while it adds the `-d` postfix for Debug configuration.

2. Exports PDB files alongside the libraries in Debug mode. This makes it possible that dependent projects have Debug symbols for Recast/Detour available, and avoids linker warnings for missing debug symbols.

Both changes should have no effects when the Microsoft compiler is not used.

### DIFF
--- a/DebugUtils/CMakeLists.txt
+++ b/DebugUtils/CMakeLists.txt
@@ -7,6 +7,7 @@ else()
 endif()
 
 add_library(RecastNavigation::DebugUtils ALIAS DebugUtils)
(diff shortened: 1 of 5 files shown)


jakobbotsch:
Thank you!

PR #448, opened 2020-10-31
Poly query API: expose distance and isOverPoly

Bromeon:
The current API `dtNavMeshQuery::findNearerestPoly()` does not allow the caller to get information about:
* whether `center` is actually _on_ a polygon
* what is the distance to the polygon (even if it's _on_ one, this can be vertical distance).

Of course it would be possible for the caller to compute these after the call, however `findNearestPoly()` has the information readily available.

I simply added an overload and didn't de-duplicate code, please let me know if you think such an API makes sense, and what would be the preferred way.

### DIFF
--- a/Detour/Include/DetourNavMeshQuery.h
+++ b/Detour/Include/DetourNavMeshQuery.h
@@ -313,24 +313,28 @@ public:
 	///@{
 
 	/// Finds the polygon nearest to the specified center point.
+	/// [opt] means the specified parameter can be a null pointer, in that case the output parameter will not be set.
+	///
 	///  @param[in]		center		The center of the search box. [(x, y, z)]
-	///  @param[in]		halfExtents		The search distance along each axis. [(x, y, z)]
+	///  @param[in]		halfExtents	The search distance along each axis. [(x, y, z)]
 	///  @param[in]		filter		The polygon filter to apply to the query.
-	///  @param[out]	nearestRef	The reference id of the nearest polygon.
-	///  @param[out]	nearestPt	The nearest point on the polygon. [opt] [(x, y, z)]
+	///  @param[out]	nearestRef	The reference id of the nearest polygon. Will be set to 0 if no polygon is found.
+	///  @param[out]	nearestPt	The nearest point on the polygon. Unchanged if no polygon is found. [opt] [(x, y, z)]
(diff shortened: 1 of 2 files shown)


Bromeon:
Any feedback on this?

jakobbotsch:
Seems reasonable to me. What is the benefit of exposing the distance? It is not exactly euclidean distance, so it might be a bit confusing, and otherwise computing it from the nearest point is trivial since the expensive part is the square root that has to be done anyway.

Bromeon:
@jakobbotsch Thanks for the feedback! Good point about the distance.

If I understand it correctly, it measures:
* the euclidean distance if the point is outside the navmesh's "airspace" (`overPoly == false`, red case)
* the vertical distance **minus walkableClimb** if it's inside (`overPoly == true`, blue case).
grafik

`walkableClimb` being taken into account means that everything that's "within climb reach" of the agent is considered zero distance. So the whole "airspace" (i.e. the prism over the polygons with height `walkableClimb`), marked grey, is closer than anything outside:
grafik

---

I'm personally only interested in the red case for now, because in RTS-style games, players often give move orders with a position slightly outside the nav-mesh, and this should be snapped to the nearest nav-mesh for some tolerance.

So yes, you're right that exposing the distance might not be necessary, as it could be inferred from `isOverPoly` and `nearestPt`. However, one would get different results without knowing the implementation, as the `walkableClimb` prism is not documented.

I'm OK either way (exposing the distance or not), however it might make sense to document how the nearest point is found. I didn't know about the prism until I looked at the code, and was indeed assuming euclidean point-polygon distance.

jakobbotsch:
I think we should remove the distance since it is not exactly the distance one might expect, so it leads to confusion. And in this case, the user can easily compute the value they want.

Bromeon:
Removed `distance` parameter. Let me know if it's good like this, or if `isOverPoly` should now simply be a default parameter (instead of an overload).

Btw, the CI failed on Mac OS with:
```
Error: sdl2: no bottle available!

You can try to install from source with:
  brew install --build-from-source sdl2```

jakobbotsch:
Thank you!
> Btw, the CI failed on Mac OS with:

Hmm yes, I will have to take a look at this at some point.

PR #449, opened 2020-10-31
Validate constraint: #polygons per tile

Bromeon:
When adding a tile to a NavMesh, ensure that the number of polygons in that tile fit in the poly ID address space.

I'm not sure if this should be done on the bits (logarithmic) level only, or if it were better if the absolute number of polygons (i.e. `dtNavMeshParams::maxPolys`) should be stored in `DetourNavMesh` and validated. However, this may be a bit harder, since we would need to sum the polygon count for each tile.

Generally, it's not very clear from the documentation how `maxPolys` is used.

### DIFF
--- a/Detour/Include/DetourNavMesh.h
+++ b/Detour/Include/DetourNavMesh.h
@@ -330,7 +330,7 @@ struct dtNavMeshParams
 	float tileWidth;				///< The width of each tile. (Along the x-axis.)
 	float tileHeight;				///< The height of each tile. (Along the z-axis.)
 	int maxTiles;					///< The maximum number of tiles the navigation mesh can contain.
-	int maxPolys;					///< The maximum number of polygons each tile can contain.
+	int maxPolys;					///< The maximum number of polygons each tile can contain. Does not influence the build step, however adding a tile with more polygons will fail.
 };
 
 /// A navigation mesh based on tiles of convex polygons.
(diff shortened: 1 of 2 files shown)


Detour/Include/DetourNavMesh.h
```diff
@@ -330,7 +330,7 @@ struct dtNavMeshParams
 	float tileWidth;				///< The width of each tile. (Along the x-axis.)
 	float tileHeight;				///< The height of each tile. (Along the z-axis.)
 	int maxTiles;					///< The maximum number of tiles the navigation mesh can contain.
-	int maxPolys;					///< The maximum number of polygons each tile can contain.
+	int maxPolys;					///< The maximum number of polygons each tile can contain. Does not influence the build step, however adding a tile with more polygons will fail.
```
jakobbotsch, line 333: Strictly speaking, Detour does not have a relevant build step, building of tiles is done by Recast.
Maybe something like the following would be clearer?

```suggestion
	int maxPolys;					///< The maximum number of polygons each tile can contain. This and maxTiles are used to calculate how many bits are needed to identify tiles and polygons uniquely.
```

Bromeon: Good point. Should it then also be added to `maxTiles`? If the user looks at the HTML documentation, not code...

jakobbotsch: Yes, that would make sense.

Bromeon: Added a commit. Let me know if you want me to squash the two instead.

jakobbotsch: No worries, I will squash when merging :)

jakobbotsch:
Thank you!

Issue #308, opened 2018-02-01
Some Problems in OffMeshConnection


MessyLove:
In the file DetourCrowd.cpp, line 1417,
is "dtCrowdAgent* ag = agents[i];"
should be "dtCrowdAgent* ag = m_agents[i];"

I was confused.

MJacred:
this can be closed due to #450

PR #450, opened 2020-11-13
fix an index error in DetourCrowd.cpp

Wisdommmmmmmm:
Fix for #308 .
The upper limit of the index of `agents` cannot be `m_maxAgents`. So I change it to `nagents`. So the index of `m_agentAnims` shoud be consistent with where it is set in line 1153.


Wisdommmmmmmm:
The upper limit of the index of `agents` cannot be `m_maxAgents`.

jakobbotsch:
Good catch, thank you!

Issue #315, opened 2018-04-27
STL in Recast & Detour


jakobbotsch:
The decision not to use STL and be very strict with dynamic memory allocations in Recast & Detour is starting to show up in a lot of places where it hurts us. For example:
#307: Can be solved easier with a simple queue
#314: Shoehorning data-structures into the existing `rcIntArray` as separate fields (actually we do this in a lot of places)
#151: Turns out `rcIntArray` was not optimized well
I am sure there are more. I suggest that we get rid of our own dynamic data structures and begin using STL instead. I think the old reasons not to use STL are largely obsolete by now: STL should be well optimized on almost any relevant platform. and the dynamic allocations we need are usually so small that the memory allocated should not be a problem.

Thoughts?

cc @grahamboree, @hymerman, @memononen

jakobbotsch:
I want to make it clear that I don't suggest drastic changes; for example, span allocation code and such should be left as-is (it has been carefully thought about to be as good as possible). It is mainly in places where there are glaringly obvious solutions to problems that we have needed to avoid for STL reasons.

hymerman:
My personal opinion is that yes, we should allow use of the STL in both Recast and Detour. That definitely doesn't mean we shouldn't be careful with memory allocation of course :) But I agree completely that there are cases where STL containers etc. will make code clearer, less robust and potentially even faster. I've never liked `rcIntArray`!!

This is something that should go towards a 2.0.0 version I guess, since it's a change in the library's dependencies.

memononen:
I've had a lot of feedback over the years thanking that Recast did not use STL.

Most game studios where I have worked have not used STL, for one reason or other. The two most common reasons have been that: 1) different implementations on different platforms have small subtle differences, which makes hard to find bugs, 2) dealing with custom allocators in STL is pain, not impossible, but just super annoying.

In some cases it makes sense to roll your own algorithm to better match your data structure/access pattern, but that is problem with all general purpose code.

I definitely went too far with repurposing the int array. A lot of tool code I worked with at that time were really sloppy with allocations and data structures, and thus slow. There were maps of vectors of sets of strings. I wanted restrict myself on dynamic allocations.

I think you should choose the way that you feel comfortable maintaining. I lightly recommend to not use STL at runtime (Detour), and if you use it in Recast, make sure it works with the custom allocators.

mbabinski-at-google:
[...]

PR #467, opened 2021-01-24
Avoid reallocs of temp vectors

glebm:
When profiling recast (as part of openmw), I found a large amount of
time in RecastMeshDetail is spent in reallocing the same vector.

This is because the `resize` method always realloc'd on grow.
Fixes this by only reallocing when needed.

Comparison for `push3` before/after second commit (Debug build though):

Before:
Screenshot from 2021-01-24 02-17-47

After:
Screenshot from 2021-01-24 02-28-14

/cc @elsid  @psi29a

### DIFF
--- a/Recast/Include/RecastAlloc.h
+++ b/Recast/Include/RecastAlloc.h
@@ -303,6 +303,7 @@ public:
 	rcIntArray(int n) : m_impl(n, 0) {}
 	void push(int item) { m_impl.push_back(item); }
 	void resize(int size) { m_impl.resize(size); }
+	void clear() { m_impl.clear(); }
 	int pop()
 	{
 		int v = m_impl.back();
--- a/Recast/Include/RecastAlloc.h
+++ b/Recast/Include/RecastAlloc.h
@@ -214,19 +214,28 @@ void rcVectorBase<T, H>::resize_impl(rcSizeType size, const T* value) {
 		destroy_range(size, m_size);
 		m_size = size;
 	} else if (size > m_size) {
-		T* new_data = allocate_and_copy(size);
-		// We defer deconstructing/freeing old data until after constructing
-		// new elements in case "value" is there.
-		if (value) {
-			construct_range(new_data + m_size, new_data + size, *value);
+		if (size <= m_cap) {
+			if (value) {
+				construct_range(m_data + m_size, m_data + size, *value);
+			} else {
+				construct_range(m_data + m_size, m_data + size);
+			}
+			m_size = size;
 		} else {
-			construct_range(new_data + m_size, new_data + size);
+			T* new_data = allocate_and_copy(size);
+			// We defer deconstructing/freeing old data until after constructing
+			// new elements in case "value" is there.
+			if (value) {
+				construct_range(new_data + m_size, new_data + size, *value);
+			} else {
+				construct_range(new_data + m_size, new_data + size);
+			}
+			destroy_range(0, m_size);
+			rcFree(m_data);
+			m_data = new_data;
+			m_cap = size;
+			m_size = size;
 		}
-		destroy_range(0, m_size);
-		rcFree(m_data);
-		m_data = new_data;
-		m_cap = size;
-		m_size = size;
 	}
 }
 template <typename T, rcAllocHint H>
--- a/Recast/Include/RecastAlloc.h
+++ b/Recast/Include/RecastAlloc.h
@@ -106,6 +106,8 @@ class rcVectorBase {
 	// Creates an array of the given size, copies all of this vector's data into it, and returns it.
 	T* allocate_and_copy(rcSizeType size);
 	void resize_impl(rcSizeType size, const T* value);
+	// Requires: min_capacity > m_cap.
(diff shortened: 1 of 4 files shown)


Bromeon:
Nice analysis! 

> Looking into it, I found lots of temporary vectors that all used resize(0), which actually deallocs memory.

It seems like [`rcVectorBase::resize_impl()` does not deallocate on shrink](https://github.com/recastnavigation/recastnavigation/blob/master/Recast/Include/RecastAlloc.h#L213-L215), just destroy the removed elements. And `rcVectorBase` is used internally by `rcIntArray`, or what am I missing?

> We also now avoid needless reallocing in RecastMeshDetail `push3`.

Good point. But I think the bigger problem is that [the `rcVectorBase::resize()` function has a very inefficient growing strategy](https://github.com/recastnavigation/recastnavigation/blob/9337e124182697de93acb656ef25766486738807/Recast/Include/RecastAlloc.h#L216-L230) -- it reallocates on every bigger resize. Even worse, it does so **even if `m_cap` is not exhausted**.

While a bit off-topic, this raises the question why is `std::vector` not used here? It gets all those things right...

jakobbotsch:
Thanks for the analysis and fix. It looks like I missed this rather unfortunate change of semantics in `resize` in #318, which definitely was NOT good for performance. `push3` was originally written the way it was when `resize` merely changed the size without doing any copying (unless reallocation was necessary). Back then, I measured that the bounds checks in the three `push` calls were significant in this hot code path (#151). I think it would be better to change `resize` back to the old (pre #318) semantics to get this win again, or add a different resize-without-reallocating function for it.

> While a bit off-topic, this raises the question why std::vector is not used here? It gets all those things right...

#315

cc @mbabinski-at-google

glebm:
Bromeon, you're right about "does not deallocate on shrink", I misread that code. I'll adjust the commit messages but will keep the `clear()` change as it's nicer code style.

I'll also fix the resize implementation.

glebm:
I've fixed `resize` and updated PR description. PTAL.

jakobbotsch:
> I've fixed resize and updated PR description. PTAL.

Thanks, can you fix `push3` again to avoid unnecessary bounds checks in `push`?

glebm:
@jakobbotsch Yes, that is indeed what I've done

Recast/Include/RecastAlloc.h
```diff
@@ -214,19 +214,28 @@ void rcVectorBase<T, H>::resize_impl(rcSizeType size, const T* value) {
...
+			if (value) {
+				construct_range(new_data + m_size, new_data + size, *value);
+			} else {
+				construct_range(new_data + m_size, new_data + size);
+			}
+			destroy_range(0, m_size);
+			rcFree(m_data);
+			m_data = new_data;
+			m_cap = size;
+			m_size = size;
```
jakobbotsch, line 237: As @Bromeon points out, this strategy is still very inefficient for `push3` since it will repeatedly allocate on every `push3`. Instead, we should grow it by doubling the capacity until it is larger than `n`. The old implementation before #318 was https://github.com/recastnavigation/recastnavigation/blob/18562383f4c5cffa0678c709049340516ebc5e40/Recast/Source/RecastAlloc.cpp#L76-L85.
Can you do something similar here?

psi29a: I agree with this, it's how I fixed inefficiencies in wildmidi. Each time we had to realloc, we just doubled the size. As th saying goes, RAM is cheap.

Bromeon:
@jakobbotsch 
> > While a bit off-topic, this raises the question why std::vector is not used here? It gets all those things right...
> 
> #315

Thank you for the link! Interesting read, even if I personally think, a lot of things have improved since the 2000s and switching to STL might actually _decrease_ maintenance efforts (this issue is a good example).

Just to understand correctly -- is no STL a decided thing then? So even if new code were to be added, we would keep handcrafting our own containers and algorithms, like the clear/grow/resize in this issue? This is not sarcastic, but rather to know where the library is heading and what contributions should look like.

jakobbotsch:
> Thank you for the link! Interesting read, even if I personally think, a lot of things have improved since the 2000s and switching to STL might actually _decrease_ maintenance efforts (this issue is a good example).
> 
> Just to understand correctly -- is no STL a decided thing then? So even if new code were to be added, we would keep handcrafting our own containers and algorithms, like the clear/grow/resize in this issue? This is not sarcastic, but rather to know where the library is heading and what contributions should look like.

I don't think anyone is saying our maintenance efforts are decreased by rolling our own basic structures, and I echo your same sentiments. But since we have people in that thread saying they use R&D in non-STL environments, I would say that yes, it is decided for now that we avoid the STL except for in the demo.

Bromeon:
Acknowledged. Thanks for clarification, and let us know in case that changes one day 🙂

Recast/Include/RecastAlloc.h
glebm: Done

jakobbotsch:
Thank you for the fix!

psi29a:
Nice job @glebm ! :)

PR #468, opened 2021-02-04
make the detourcrowd go across the offmeshlink parabolic

hbzzhtt:
The RecastDemo example is very elegant, but the crowd moves in a straight line when it crosses the offmesh link. In real life, the parabola moves in a more realistic way, so I changed the crowd's moving pattern to a parabola.

### DIFF
--- a/Detour/Include/DetourCommon.h
+++ b/Detour/Include/DetourCommon.h
@@ -121,7 +121,21 @@ inline void dtVlerp(float* dest, const float* v1, const float* v2, const float t
 	dest[1] = v1[1]+(v2[1]-v1[1])*t;
 	dest[2] = v1[2]+(v2[2]-v1[2])*t;
 }
-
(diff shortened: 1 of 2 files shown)


grahamboree:
Thanks for this PR!  I agree that parabolic movement looks better in demos like this than straight lerping movement.  

Unfortunately, parabolic movement is much harder for an animator to author an animation for.  Additionally not all off-mesh-links are parabolic movement.  Sometimes links like this are used for things like doorways that can be opened or closed, rather than gaps that must be jumped over.

The lerping happening here is really just about expressing a logical position change, rather than trying to procedurally control a character across a nav link.  A lot of what the "right thing to do" for a nav link is entirely dependent on the game as well as the specific nav link, so it's impossible to find a universal solution for everyone.  For this reason it's best to keep the simplest approach (lerp) and leave implementation of better looking movement to the user.

PR #470, opened 2021-02-20
fix:rcBuildLayerRegions missing areaType

shineTeam7:
fix:rcBuildLayerRegions missing areaType


jankrassnigg:
This fix is absolutely vital to get area IDs working correctly. Here is what happens without it:

grafik

The black blocks use a different area ID and should show up in red rather than blue. As you can see, the blue area ID takes over in many places. The dark blue grid lines are tile borders, so that's why the area ID is correct for half the black block, but not the other half, because a tile border separates them.

With the fix, the area IDs work as expected:

grafik

It would be great, if this PR could be merged to the main branch.

PR #471, opened 2021-02-20
duDisplayList: Fix -Wsuggest-override warnings

glebm:
Fixes warnings such as this one:

```
DebugUtils/Include/DebugDraw.h:209:15: warning: ‘virtual void duDisplayList::depthMask(bool)’ can be marked override [-Wsuggest-override]
  209 |  virtual void depthMask(bool state);
      |               ^~~~~~~~~
```

### DIFF
--- a/DebugUtils/Include/DebugDraw.h
+++ b/DebugUtils/Include/DebugDraw.h
@@ -206,11 +206,11 @@ class duDisplayList : public duDebugDraw
 public:
 	duDisplayList(int cap = 512);
 	~duDisplayList();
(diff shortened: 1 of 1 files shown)


akortunov:
I suppose that R&D maintains compatibility with legacy C++ code, so this solution will not work - an `override` keyword was introduced only in C++11.

grahamboree:
Appreciate the contribution, but as was previously mentioned, we aim to be maximally compatible, and as such we stick to C++98 features.  The `override` keyword was added in C++11.

PR #473, opened 2021-03-04
Install .pdb files only on windows

elsid:
These files are not generated on Linux that leads to an error:
```
CMake Error at DebugUtils/cmake_install.cmake:60 (file):
  file INSTALL cannot find
  "/home/elsid/dev/recastnavigation/build/gcc/debug/DebugUtils/DebugUtils-d.pdb":
  No such file or directory.
Call Stack (most recent call first):
  cmake_install.cmake:47 (include)
```

### DIFF
--- a/DebugUtils/CMakeLists.txt
+++ b/DebugUtils/CMakeLists.txt
@@ -33,4 +33,6 @@ install(TARGETS DebugUtils
 file(GLOB INCLUDES Include/*.h)
 install(FILES ${INCLUDES} DESTINATION
     ${CMAKE_INSTALL_INCLUDEDIR}/recastnavigation)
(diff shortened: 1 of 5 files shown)


Bromeon:
Thanks for the PR 🙂 

I think you could even use [the `MSVC` variable](https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/Useful-Variables#system-compiler-information) instead of `WIN32`, as .pdb files are only generated by the Microsoft Compiler, not e.g. MinGW.

elsid:
Replaced WIN32 by MSVC.

jakobbotsch:
Thank you!

PR #474, opened 2021-03-04
Add version and pkgconfig

psi29a:
A more module approach to versioning. This will help to get this suite of libraries into Debian/Ubuntu, having its own recastnavigation.pc (pkgconfig) file.

/cc @glebm


CMakeLists.txt
```diff
@@ -4,7 +4,11 @@ project(RecastNavigation)
 
 # lib versions
 SET(SOVERSION 1)
-SET(VERSION 1.0.0)
```
glebm, line 7: Can be simplified a bit:

```cmake
set(LIB_VERSION 1.5.1)
string(REPLACE LIB_VERSION "." "," LIB_VERSION_NUM)
set(LIB_VERSION_NUM, "${LIB_VERSION_NUM},0")
```

psi29a: Thanks, yes it needn't be so verbose. It needn't be exactly like OpenAL-Soft ;)

jakobbotsch:
Thank you!

PR #475, opened 2021-03-05
Fix MacOS builds on Travis

psi29a:
use defaults where possible, simplify .travis.yml and fix macos builds while we are at it

/cc @glebm


jakobbotsch:
That's great, the broken build has been annoying me a while :) Thank you for taking care of it!

psi29a:
Gern geschehen! :)

PR #476, opened 2021-04-04
Fix incorrect rasterization at tile borders

ppiastucki:
Currently the coordinates are simply clamped to the allowed range producing incorrect rasterization at tile borders.  The problem can be easily reproduced in RecastDemo. 
Since the polygon is not cut at the border of the tile when the first row (y=0) is processed an incorrect range of x coordinates is computed based on all vertices of the original polygon including the ones in the "upper" part lying completely outside of the tile. This results in the rasterization of the **red** voxels in the picture.

Since the x coordinate is simply clamped without any checks, a voxel at the border may be rasterized even though the whole polygon row lies completely outside of the tile resulting in the rasterization of the **green** voxels in the picture.

The suggested fix is to:
* clip the polygon at the tile border by starting the scan at y=-1 and x=-1 respectively and
* disregard x coordinates outside of the tile before clamping.

incorrect_rasterization2

_Before_
Before
_After_
After


Recast/Source/RecastRasterization.cpp
```diff
@@ -264,7 +264,8 @@ static bool rasterizeTri(const float* v0, const float* v1, const float* v2,
 	// Calculate the footprint of the triangle on the grid's y-axis
 	int y0 = (int)((tmin[2] - bmin[2])*ics);
 	int y1 = (int)((tmax[2] - bmin[2])*ics);
-	y0 = rcClamp(y0, 0, h-1);
+	// use -1 rather than 0 to cut the polygon properly at the start of the tile
+	y0 = rcClamp(y0, -1, h-1);
```
gre-42, line 268: This also fixes a bug when h=0. However, the line below still contains this bug.

onehundredfeet:
I've seen this a lot.  Please integrate.

grahamboree:
Thank you for this PR and for all the helpful info about the problem!  The diagram is excellent. :) Apologies this took so long to integrate.  I agree this is a pretty serious issue with tile rasterization that should be fixed.  Thankfully these stray voxels usually get tossed out later in the process because the region they create is too small.  

Here's a more precise set of repro steps:
1. Run RecastDemo and open the Tile Mesh Sample
2. Choose the nav_test.obj input mesh
3. Set the tile size to 112
4. Check "Keep Intermediate Results"
5. Build the navmesh
6. Select Voxels draw mode, and click on the tile shown in the screenshot 

I dug into the problem a bit more.  It seems to occur when rasterizing a triangle whose bounding box overlaps the heightfield (so it's not immediately discarded) and whose x and z extents overlap the minimum heightfield x and z respectively.

I've added a minimal repro test case with the name "rcRasterizeTriangle overlapping bb but non-overlapping triangle" to our test suite that tests this issue.  I verified that this test fails with the current state of the rasterization code, and passes with this fix.

ppiastucki:
Thank you!

PR #479, opened 2021-04-16
Fastlz fix for -Wno-class-memaccess' is valid for C++/ObjC++ but not for C

Dante-666:
I am on Arch Linux and using GCC 10.2.0

While building RecastDemo using instructions provided, I am getting this error

Creating obj/Debug/RecastDemo
fastlz.c
ChunkyTriMesh.cpp
ConvexVolumeTool.cpp
CrowdTool.cpp
Filelist.cpp
cc1: error: command-line option '-Wno-class-memaccess' is valid for C++/ObjC++ but not for C [-Werror]

I think adding this filter works out well. Please let me know if you have some changes to suggest or if this is too platform specific.


grahamboree:
Thanks for this!  Looks like there's some knock-on failures in the Windows build now though.  https://ci.appveyor.com/project/recastnavigation/recastnavigation/builds/38741800

grahamboree:
This was fixed in #573

PR #504, opened 2021-12-06
Adds .os to gitignore

EamonnMR:
Compiling in some contexts (in my case: Ubuntu, and godotdetour) will output .os files. Could be that someone typo'd .so somewhere. https://github.com/TheSHEEEP/godotdetour

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 *.pdb
 *.so
 *.idb
+*.os
 
 ## Linux exes have no extension
 RecastDemo/Bin/RecastDemo


grahamboree:
Thanks for this!  What .os files are getting generated out of curiosity?  Maybe there's something wrong with one of the build scripts and knowing what files you're trying to ignore might help track it down.

grahamboree:
Since it's unclear what files are being generated, there's no reports of this happening to anyone else, and there's no .so file extensions typo'd in the recast build pipeline, I'm going to conclude that is coming from something in godotdetour.

PR #513, opened 2022-03-16
Fixed RecastDemo crash in macOS 10.11+

Prin-E:
- Set SDL hints to always use OpenGL render driver (instead of Metal render driver by default)
- Compile error fix in Xcode 13 (replaced <RecastAssert.h> to "RecastAssert.h")


grahamboree:
LGTM!  Merged.  Thanks!

PR #519, opened 2022-04-19
Fix crash on load

jinq0123:
Reproduce the crash:
1. Choose Sample -> Solo Mesh
2. Choose Mesh -> nav_test.obj
3. Build
4. Save
5. Load
6. Click on the map

Call stack:
```
>	RecastDemo.exe!dtNavMesh::getTileAndPolyByRef(const unsigned int ref, const dtMeshTile * * tile, const dtPoly * * poly) 行 1203	C++
 	RecastDemo.exe!duDebugDrawNavMeshPoly(duDebugDraw * dd, const dtNavMesh & mesh, unsigned int ref, const unsigned int col) 行 445	C++
 	RecastDemo.exe!NavMeshTesterTool::handleRender() 行 1070	C++
 	RecastDemo.exe!Sample_SoloMesh::handleRender() 行 342	C++
 	RecastDemo.exe!SDL_main(int __formal, char * * __formal) 行 493	C++
 	RecastDemo.exe!main_getcmdline() 行 82	C
 	[外部代码]
```

### DIFF
--- a/RecastDemo/Source/Sample_SoloMesh.cpp
+++ b/RecastDemo/Source/Sample_SoloMesh.cpp
@@ -102,6 +102,8 @@ void Sample_SoloMesh::handleSettings()
 		dtFreeNavMesh(m_navMesh);
 		m_navMesh = Sample::loadAll("solo_navmesh.bin");
 		m_navQuery->init(m_navMesh, 2048);
+		if (m_tool)
+			m_tool->init(this);
 	}
 
 	imguiUnindent();
--- a/RecastDemo/Source/Sample_TempObstacles.cpp
+++ b/RecastDemo/Source/Sample_TempObstacles.cpp
@@ -930,6 +930,8 @@ void Sample_TempObstacles::handleSettings()
 		dtFreeTileCache(m_tileCache);
 		loadAll("all_tiles_tilecache.bin");
 		m_navQuery->init(m_navMesh, 2048);
+		if (m_tool)
+			m_tool->init(this);
 	}
 
 	imguiUnindent();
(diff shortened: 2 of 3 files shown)


jinq0123:
Sorry. I tested again and found it non-reproducible. 

Then I found a way to reproduce it:
```
dtNavMesh::~dtNavMesh()
{
        ...
	dtFree(m_tiles);
+	m_tiles = 0;
}
```

grahamboree:
I added `m_tiles = 0;` to line 222 in `DetourNavMesh.cpp` and followed your original repro steps but wasn't able to get the demo app to crash.  Can you post detailed steps to exactly reproduce the crash?

jinq0123:
"Load" will free the old navmesh. `m_tool->init()` should be called to set new navmesh in m_tool.
```
	if (imguiButton("Load"))
	{
		dtFreeNavMesh(m_navMesh);
		m_navMesh = Sample::loadAll("solo_navmesh.bin");
		m_navQuery->init(m_navMesh, 2048);
	}
```
If you didn't crash, the old m_navMesh must equal to the new m_navMesh. 

I think changing the order of free() and new() will get you to the crash:
```
	if (imguiButton("Load"))
	{
-		dtFreeNavMesh(m_navMesh);
+		auto oldNavMesh = m_navMesh;

		m_navMesh = Sample::loadAll("solo_navmesh.bin");
		m_navQuery->init(m_navMesh, 2048);

+		assert(oldNavMesh != m_navMesh);
+		dtFreeNavMesh(oldNavMesh);
	}
```

grahamboree:
Sorry, I've tried every combination of generating, saving, loading, and switching meshes and I haven't been able to reproduce the crash.  Can you post the precise steps you're using to reproduce the issue?

PR #520, opened 2022-04-24
Fix distance in rcErodeWalkableArea()

jinq0123:
Wrong distance results less eroded, which will cause the collision with the boundary.

### DIFF
--- a/Recast/Source/RecastArea.cpp
+++ b/Recast/Source/RecastArea.cpp
@@ -50,9 +50,12 @@ bool rcErodeWalkableArea(rcContext* ctx, int radius, rcCompactHeightfield& chf)
 		ctx->log(RC_LOG_ERROR, "erodeWalkableArea: Out of memory 'dist' (%d).", chf.spanCount);
 		return false;
 	}
+
 	const double DIST_MAX = w + h + radius;
 	const double DIST_NEIGHBOUR = 1; // distance between neighbour cells
 	const double DIST_DIAGONAL = sqrt(2); // distance between diagonal neighbour cells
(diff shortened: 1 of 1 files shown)


jinq0123:
I think the error will emerge if the cell size is less than 3/2-sqrt(2).

grahamboree:
I'm not sure this is the best solution here.  

The problem you're getting at seems to be related to the inaccuracy of using 3 and 2 as estimates of straight line and diagonal distances.  2 is accurate but 3 is an approximation (`2*sqrt(2)`).  The inaccuracy between this approximate measurement grows as the distance grows, so you're right to observe that it's relevant when the voxel size is small and the diagonal voxel distances to edges are large.
image

The rationale behind this decision is that it makes distance checks significantly faster and simpler.  Using doubles here removes the inaccuracy but also uses 8x as much memory and is slower.

Because the distance values are capped at 255, the cases where this becomes relevant are when the diagonal voxel distance is between 17 and 90 cells.  In those distances, the approximated value is off by at least 1 diagonal cell distance from the actual value.  The impact on issues like colliding with walls will then depend on the ratio of this distance error with the radius of your agent.  The closer the agent radius is to the size of the voxel, the more impactful this becomes.

In general Recast is very sensitive to voxel parameters, and users tend to turn down voxel size with the goal of improving the fidelity of the generated navmesh.  The better way to achieve this is to use a tiled mesh.  Tiled meshes let you set the voxel size smaller while also ensuring that the distance checks in the erosion step aren't huge, since those distances are limited to within the mesh tile.

PR #525, opened 2022-06-01
-Wunused-but-set-variable flagged variables removed

bpswenson:
Was causing o3de to not build since it had -Werror and -Wunused-but-set-variable


grahamboree:
lgtm

PR #554, opened 2022-10-09
Correction the comment of dtVperp2D

ZA139:
In Detour/Include/DetourCommon.h:322
`/// @return The dot product on the xz-plane.`
The return value should be the perp dot product rather than the dot product.


grahamboree:
Thanks for this!  Looks like your fix has a typo though

ZA139:
> Thanks for this! Looks like your fix has a typo though

Sry about that. Already fixed it. :)

PR #560, opened 2022-10-12
fix: dtNavMeshQuery::findRandomPoint should not fail because of getPo…

### LINKED ISSUE #556: dtNavMeshQuery::findRandomPoint got failed sometimes
https://github.com/recastnavigation/recastnavigation/blob/314e94b692fe6c5df1de703527c8774a473526b7/Detour/Source/DetourNavMeshQuery.cpp#L301-L307

In some case, the `pt` calculated by `dtRandomPointInConvexPoly` would be exactly at right edge (+x) of this poly, e.g:
image
And it failed in `getPolyHeight` after, because `dtPointInPolygon` said this point is not in the triangle.

I know the last part is a feature of `dtPointInPolygon`, but it is really weird if I just want a random position and got `DT_INVALID_PARAM`.

EfveZombie:
fix #556


Detour/Source/DetourNavMeshQuery.cpp
```diff
@@ -481,23 +482,24 @@ dtStatus dtNavMeshQuery::findRandomPointAroundCircle(dtPolyRef startRef, const f
...
 	dtStatus stat = getPolyHeight(randomPolyRef, pt, &h);
-	if (dtStatusFailed(status))
-		return stat;
-	pt[1] = h;
-	
+	if (dtStatusSucceed(stat)) {
+		// otherwise pt is on poly edge
+		pt[1] = h;
+	}
+
```
jakobbotsch, line 498: I think this should be switched to `closestPointOnPoly` instead. Also, can you please undo the whitespace changes (or submit them as a separate fix)? Thanks!

EfveZombie: Got it

jakobbotsch:
Thank you!

PR #571, opened 2022-10-26
Move FastLZ to it's own project in premake5 config

grahamboree:
...so we don't get errors about certain flags being C++ only.  Specifically we get this error when compiling on Ubuntu with gcc:

```cc1: error: command-line option ‘-Wno-class-memaccess’ is valid for C++/ObjC++ but not for C [-Werror]```

Moving FastLZ to a separate project removes this issue.

This isn't a terribly aesthetic solution (yet another project in the workspace) but it works and is clean.

### DIFF
--- a/RecastDemo/premake5.lua
+++ b/RecastDemo/premake5.lua
@@ -116,6 +116,17 @@ project "Recast"
 		"../Recast/Source/*.cpp" 
 	}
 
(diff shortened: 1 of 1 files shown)


grahamboree:
Nevermind.  #479 is a better solution for this

Issue #413, opened 2019-11-23
RecastDemo doesn't build


jdestefx:
Centos 7.6

[root@GD-4105-SVR gmake]# make
==== Building DebugUtils (debug) ====
Creating obj/Debug/DebugUtils
DebugDraw.cpp
DetourDebugDraw.cpp
RecastDebugDraw.cpp
RecastDump.cpp
Creating lib/Debug
Linking DebugUtils
==== Building Detour (debug) ====
Creating obj/Debug/Detour
DetourAlloc.cpp
cc1plus: error: -Werror=class-memaccess: no option -Wclass-memaccess
make[1]: *** [obj/Debug/Detour/DetourAlloc.o] Error 1
make: *** [Detour] Error 2

ash1215:
I got the same error in Ubuntu 18.04
Though deleting the code " -Wno-error=class-memaccess " from Detour.make helped me solve this error. But I don't know if deleting is a good idea. Anyone has any other solution, please let me know.

Issue #403, opened 2019-08-20
Unable to compile on ubuntu 18.04


ndeshp2s:
Hello,

I am trying to build the RecastDemo on Ubuntu 18.04. However, I get the following error when I run make in Build/gmake:
`==== Building DebugUtils (debug) ====
==== Building Detour (debug) ====
DetourAlloc.cpp
cc1plus: error: -Werror=class-memaccess: no option -Wclass-memaccess
Detour.make:133: recipe for target 'obj/Debug/Detour/DetourAlloc.o' failed
make[1]: *** [obj/Debug/Detour/DetourAlloc.o] Error 1
Makefile:46: recipe for target 'Detour' failed
make: *** [Detour] Error 2`

How can solve this issue?

Thankyou in advance.

aurthes6220:
I have met the same situation. I need someone's help too.

Dig-Doug:
As a temporary workaround, I was able to build by removing the compiler flag:

```bash
cd RecastDemo
sed -i -e 's/-Wno-error=class-memaccess//g' premake5.lua
# Run premake, etc.
```

PR #573, opened 2022-10-29
Disable class-memaccess error to fix builds on Ubuntu with newer versions of GCC

### LINKED ISSUE #497: Building error: -Wno-class-memaccess, using g++ 9.3.0 on Ubuntu 20.04
I was able to successfully build if I went through Recastdemo.make and Detour.make and removed all the references to -Wno-class-memaccess, but I thought I should post this here in case there's a better way to build it without messing with those files, or in case this is a bug.

```
==== Building DebugUtils (debug) ====
Creating obj/Debug/DebugUtils
DebugDraw.cpp
DetourDebugDraw.cpp
RecastDebugDraw.cpp
RecastDump.cpp
Creating lib/Debug
Linking DebugUtils
==== Building Detour (debug) ====
Creating obj/Debug/Detour
DetourAlloc.cpp
DetourAssert.cpp
DetourCommon.cpp
DetourNavMesh.cpp
DetourNavMeshBuilder.cpp
DetourNavMeshQuery.cpp
DetourNode.cpp
Linking Detour
==== Building DetourCrowd (debug) ====
Creating obj/Debug/DetourCrowd
DetourCrowd.cpp
DetourLocalBoundary.cpp
DetourObstacleAvoidance.cpp
DetourPathCorridor.cpp
DetourPathQueue.cpp
DetourProximityGrid.cpp
Linking DetourCrowd
==== Building DetourTileCache (debug) ====
Creating obj/Debug/DetourTileCache
DetourTileCache.cpp
DetourTileCacheBuilder.cpp
Linking DetourTileCache
==== Building Recast (debug) ====
Creating obj/Debug/Recast
Recast.cpp
RecastAlloc.cpp
RecastArea.cpp
RecastAssert.cpp
RecastContour.cpp
RecastFilter.cpp
RecastLayers.cpp
RecastMesh.cpp
RecastMeshDetail.cpp
RecastRasterization.cpp
RecastRegion.cpp
Linking Recast
==== Building RecastDemo (debug) ====
Creating obj/Debug/RecastDemo
fastlz.c
cc1: error: command line option ‘-Wno-class-memaccess’ is valid for C++/ObjC++ but not for C [-Werror]
cc1: all warnings being treated as errors
make[1]: *** [RecastDemo.make:150: obj/Debug/RecastDemo/fastlz.o] Error 1
make: *** [Makefile:69: RecastDemo] Error 2

```

grahamboree:
Based on SgtVincent's suggestion here: https://github.com/recastnavigation/recastnavigation/issues/497#issuecomment-1293318731

Demotes the warning from an error (due to warnings-as-errors) back to a warning.

Fixes #497

### DIFF
--- a/RecastDemo/premake5.lua
+++ b/RecastDemo/premake5.lua
@@ -76,7 +76,7 @@ project "Detour"
 	-- linux library cflags and libs
 	filter {"system:linux", "action:gmake"}
 		buildoptions {
-			"-Wno-class-memaccess"
+			"-Wno-error=class-memaccess"
 		}
 
 project "DetourCrowd"
@@ -155,7 +155,7 @@ project "RecastDemo"
 			"`pkg-config --cflags gl`",
 			"`pkg-config --cflags glu`",
 			"-Wno-ignored-qualifiers",
-			"-Wno-class-memaccess"
+			"-Wno-error=class-memaccess"
 
 		}
 		linkoptions { 


jakobbotsch:
Doesn't this reintroduce #403 and #413?

jakobbotsch:
The right fix for this should be something like #479, the compiler is complaining that we are passing class-related warning disables to C files.

grahamboree:
This in combination with #574 fixes compilation on Ubuntu 22.04 using gcc 11.3.0.

This seemed like the least intrusive fix for this issue as it just demotes this back to a warning.  The warning is harmless (it's just ignoring the flag) but I agree the ideal solution is something like #479 to remove the warning.  That specific PR had issues with Windows builds though, so if we can sort that out I'm all for it.  I probably should have left that PR open...

jakobbotsch:
> This seemed like the least intrusive fix for this issue

Well, since this is a revert of #432 I assume it reintroduces build errors on other GCC versions, so it just moves the problem around. Quoting that PR:

> -Wno-error does not ignore unknown warnings on old GCC versions causing
problems when using it for new warnings. On the other hand, -Wno-X will
ignore unknown warnings, so we can use this instead.

I.e. here is a godbolt that shows the behavior: https://godbolt.org/z/qGfen9781

GCC 7 is still the default on Ubuntu 18, so Recast & Detour no longer compiles on Ubuntu 18 with this change, I think.

grahamboree:
Ah yeah you're totally right.  Didn't know about #432!  Okay it's pretty clear that #479 is the best way forward.  I'll see if I can get that change to play nice with Windows.

elsid:
Could you just merge https://github.com/recastnavigation/recastnavigation/pull/480 to actually fix the warning rather than trying to hide it? This is the original compiler report:
```
Detour/Source/DetourNavMesh.cpp: In member function ‘dtStatus dtNavMesh::init(const dtNavMeshParams*)’:
Detour/Source/DetourNavMesh.cpp:243:15: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct dtMeshTile’ with no trivial copy-assignment [-Wclass-memaccess]
  243 |         memset(m_tiles, 0, sizeof(dtMeshTile)*m_maxTiles);
      |         ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
And it's valid since `dtMeshTile` has user defined copy assignment operator for no good reason as explained in the mentioned PR in detail.

PR #575, opened 2022-10-30
Install CMake config file to allow find_package()

SpaceIm:
https://github.com/recastnavigation/recastnavigation/pull/358 has added ALIAS targets, but the title was misleading, ALIAS targets are only for add_subdirectory(), not find_package().

This PR installs a CMake config file containing imported targets of recastnavigation. These targets are consistent with ALIAS one: `RecastNavigation::DebugUtils`, `RecastNavigation::Detour`, `RecastNavigation::DetourCrowd`, `RecastNavigation::DetourTileCache`, `RecastNavigation::Recast`.

Thus, one can find an installation of recastnavigation and link targets, for example:

```cmake
find_package(recastnavigation REQUIRED)
target_link_libraries(foo PRIVATE RecastNavigation::Recast RecastNavigation::Detour)
```

```
mdkir build & cd build
cmake .. -DCMAKE_PREFIX_PATH=<path/to/recastnavigation/install/dir>
```

(it can also work against build tree of recastnavigation)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,7 @@ if(MSVC AND BUILD_SHARED_LIBS)
     set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 endif()
 
+include(CMakePackageConfigHelpers)
 include(GNUInstallDirs)
 
 configure_file(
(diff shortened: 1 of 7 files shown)


grahamboree:
Appreciated, thanks!

Issue #550, opened 2022-10-07
Migrate CI to GitHub Actions


grahamboree:
There's a lot of upsides to using GitHub actions.  We should migrate our existing CI setup with Travis and AppVeyor to use Actions instead.

grahamboree:
Fixed in #580

Issue #577, opened 2022-11-08
CMake SDL2 include paths missing on macOS


grahamboree:
## Setup:
m1 MacBook Pro
macOS 13.0
SDL2.framework installed to /Library/Frameworks

## Repro:
build with the following commands run from the root of a clean copy of the repo
```
cmake -B ./build -DCMAKE_BUILD_TYPE=Debug
cmake --build ./build --config Debug
```

results in:
```
...
[ 60%] Building CXX object RecastDemo/CMakeFiles/RecastDemo.dir/Source/ConvexVolumeTool.cpp.o
In file included from /Users/grahampentheny/workspace/recastnavigation/RecastDemo/Source/ConvexVolumeTool.cpp:24:
/Library/Frameworks/SDL2.framework/Headers/SDL.h:32:10: fatal error: 'SDL2/SDL_main.h' file not found
#include <SDL2/SDL_main.h>
         ^~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [RecastDemo/CMakeFiles/RecastDemo.dir/Source/ConvexVolumeTool.cpp.o] Error 1
make[1]: *** [RecastDemo/CMakeFiles/RecastDemo.dir/all] Error 2
make: *** [all] Error 2
```

I suspect this line in `RecastDemo/CMakeLists.txt` is to blame, though I'm not sure how frameworks are handled by cmake.
```
include_directories(Contrib/SDL/include)
```

---

## Fix Requirements:
We ultimately need the cmake script to invoke clang with the `-framework SDL2` and `-I/Library/Frameworks/SDL2.framework/Headers` options.  

It would also be nice to be able to specify the SDL framework path (via `-F` clang flag) so it's not just hardcoded to `/Library/Frameworks`.  Specifying the framework path like this is useful for CI processes that might not want to or be able to muck about in `/Library`.

elsid:
There is `SDL2_ROOT_DIR` variable to provide a path to SDL2. Run CMake using it to make CMake find the library:
```
cmake -D SDL2_ROOT_DIR=<path> <...>
```

I have no way to test this on macOS but I see [some support](https://github.com/recastnavigation/recastnavigation/blob/b0d71bed92a89c0b4ad63ca4278d12c63d8d6eb4/RecastDemo/cmake/FindSDL2.cmake#L82-L102) for the frameworks in the FindSDL2.cmake.

[...]

PR #580, opened 2022-11-12
Github Workflows based CI system to replace Appveyor and Travis

grahamboree:
The goal here is to replace the current (slightly broken) Travis + Appveyor setup with something that covers more cases and is a bit better supported and easier to maintain.  This hopefully helps us catch cross-platform issues quicker.  For example, the linux and clang compilation issues that have existed for a while could have been caught if we were building those targets in CI.

This adds a build script that builds the following configurations for every repo commit and PR:
- macOS, premake, Debug
- macOS, premake, Release
- linux, premake, gcc, Debug
- linux, premake, gcc, Release
- linux, premake, clang, Debug
- linux, premake, clang, Release
- linux, cmake, Debug
- linux, cmake, Release
- windows, premake, vs2019, Debug
- windows, premake, vs2019, Release
- windows, premake, vs2022, Debug
- windows, premake, vs2022, Release
- windows, cmake, vs2019, Debug
- windows, cmake, vs2019, Release
- windows, cmake, vs2022, Debug
- windows, cmake, vs2022, Release

It also builds and runs the catch tests executable in the following configurations.  A failed test will fail the build.
- macOS, premake, Debug
- linux, premake, clang, Debug
- windows, premake, vs2022, Debug

It doesn't currently build cmake on macOS because there is a one blocker there (#577).  We could also add additional builds like cmake with both clang and gcc on linux, but for now this is much better coverage than we have currently.


grahamboree:
Fixes #550

PR #592, opened 2022-12-14
Cleanup Rasterization Code

grahamboree:
Mostly just improving variable names for clarity.

Removed a redundant base-case check in AddSpan that resulted in a small but measurable perf improvement due to removing a branch.  Also tightened up the bounding box overlap check so it early-outs faster, which also has a small perf improvement to rasterization times.

Updated docstrings for user-facing functions.

### DIFF
--- a/Recast/Include/Recast.h
+++ b/Recast/Include/Recast.h
@@ -842,11 +842,14 @@ void rcClearUnwalkableTriangles(rcContext* ctx, const float walkableSlopeAngle, 
 ///  @param[in]		area			The area id of the span. [Limit: <= #RC_WALKABLE_AREA)
 ///  @param[in]		flagMergeThr	The merge theshold. [Limit: >= 0] [Units: vx]
 ///  @returns True if the operation completed successfully.
-bool rcAddSpan(rcContext* ctx, rcHeightfield& hf, const int x, const int y,
-			   const unsigned short smin, const unsigned short smax,
-			   const unsigned char area, const int flagMergeThr);
+bool rcAddSpan(rcContext* ctx, rcHeightfield& hf,
+	           int x, int y,
+               unsigned short smin, unsigned short smax,
+               unsigned char area, int flagMergeThr);
 
-/// Rasterizes a triangle into the specified heightfield.
+/// Rasterizes a single triangle into the specified heightfield.
+///
+/// Calling this for each triangle in a mesh is less efficient than calling rcRasterizeTriangles
 ///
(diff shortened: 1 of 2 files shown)


Recast/Source/RecastRasterization.cpp
```diff
@@ -23,433 +23,532 @@
...
+/// Check whether two bounding boxes overlap
+///
+/// @param[in]	aMin	Min axis extents of bounding box A
+/// @param[in]	aMax	Max axis extents of bounding box A
+/// @param[in]	bMin	Min axis extents of bounding box B
+/// @param[in]	bMax	Max axis extents of bounding box B
+/// @returns true if the two bounding boxes overlap.  False otherwise.
+static bool overlapBounds(const float* aMin, const float* aMax, const float* bMin, const float* bMax)
 {
-	bool overlap = true;
```
memononen, line 28: This came from a project where it lead to branchless test.

Using `const float*` leads to a lot of branches, but a version using simple `vec3` struct gets vectorized:
https://godbolt.org/z/oaKPMqW7x

I think that could lead a good savings even if used just internally.

grahamboree: Oh nice! I was hoping to do something like this already for clarity but it makes a ton of sense that it'd also improve perf since it gives much more information to the optimizer and changes things to value semantics, which eliminates the possibility of pointer shadowing.  Definitely an improvement we should make. 💯

grahamboree: I'll do some more profiling but given that the original code here was branchless it's probably better to keep that over this version.

memononen: The changed are that the original code was either using `Vec3` of sorts or `float[3]` which would allow the compiler to vectorize it. I think your version has should be better with `float*`

Recast/Include/Recast.h
```diff
@@ -822,81 +825,110 @@ void rcClearUnwalkableTriangles(rcContext* ctx, const float walkableSlopeAngle,
 								const int* tris, int nt, unsigned char* areas); 
 
 /// Adds a span to the specified heightfield.
+/// 
+/// The span addition can be set to favor flags. If the span is merged to
+/// another span and the new @p smax is within @p flagMergeThr units
```
ppiastucki, line 830: Old parameter names are still used here and below: smax, flagMergeThr, Limit: < @p smax

grahamboree: Fixed in 86383fc

PR #593, opened 2022-12-14
Supports offmesh links between non-neighbor tiles

alpaca2333:
In some cases, a (relatively) long range nav link is required, such as making NPC moving through a zipline.
However, recast will only create links between tiles that are neighbors. 
So I implemented the feature and had it tested in my game, and in recast app.

It is a small enhancement.

### DIFF
--- a/Detour/Source/DetourNavMesh.cpp
+++ b/Detour/Source/DetourNavMesh.cpp
@@ -1347,7 +1347,6 @@ dtStatus dtNavMesh::removeTile(dtTileRef ref, unsigned char** data, int* dataSiz
 	{
 		for (const auto &farLinks: iter->second)
 		{
-			
 			nneis = getTilesAt(farLinks.x, farLinks.y, neis, MAX_NEIS);
 			for (int j = 0; j < nneis; ++j)
 			{
--- a/Detour/Source/DetourNavMesh.cpp
+++ b/Detour/Source/DetourNavMesh.cpp
@@ -136,6 +136,16 @@ inline void freeLink(dtMeshTile* tile, unsigned int link)
 }
 
 
+void dtNavMesh::registerFarLinks(int x1, int y1, int x2, int y2, int side)
+{
+	auto iter = m_farLinks.find(std::make_pair(x1, y1));
+	if (iter == m_farLinks.end())
+	{
+		m_farLinks.emplace(std::make_pair(x1, y1), dtFarLinksSet());
+	}
+	m_farLinks[std::make_pair(x1, y1)].emplace(dtNonNeighborLink{x2, y2, side});
+}
+
 dtNavMesh* dtAllocNavMesh()
 {
 	void* mem = dtAlloc(sizeof(dtNavMesh), DT_ALLOC_PERM);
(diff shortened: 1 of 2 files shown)


Detour/Source/DetourNavMesh.cpp
```diff
@@ -518,7 +519,30 @@ void dtNavMesh::connectExtOffMeshLinks(dtMeshTile* tile, dtMeshTile* target, int
...
+		calcTileLoc(targetPos, &tx, &ty);
+		if (abs(tx - tile->header->x) <= 1 && abs(ty - tile->header->y) <= 1)
+		{
+			continue;
+		}
+		int ntiles = getTilesAt(tx, ty, tiles, MAX_TILES);
+		for (int j = 0; j < ntiles; ++j)
+		{
+			connectExtOffMeshLinks(tile, tiles[j], selfCon->side);
+			connectExtOffMeshLinks(tiles[j], tile, dtOppositeTile(selfCon->side));
```
memononen, line 543: This does not work correctly. Connections are stored in the tile where their start points are.

If you have connection from tile (1,1) to tile (5,1), and first add tile (1,1) and then tile (5,1) the off-mesh link will not get connected, because tile (5,1) was not around when the first tile was added.

Similar issue arises when tiles (5,1) is removed before tile (1,1).

This is the reason why only neighbour tiles are considered for off-mesh links, as we can do the reciprocal add/remove as there aren't that many tiles to consider.

alpaca2333: @memononen thanks for your reply!
the `connectExtOffMeshLinks` function will connect all links from `target` to this `tile`, and it will be called both way. So in my opinion it is ok. The original use of `connectExtOffMeshLinks` between neighbors is already redundant.

what do you see?

alpaca2333: @memononen also, the newly-added function will look through all links and accurately find out which tiles will be linked to.
there is nothing different between connecting (1, 1) to (5, 1)  and connecting (1, 1) with its neighbors.

what i do is changing 'considering only neighbors' to 'considering all tiles which is calculated by link's position'

alpaca2333: however, what you say about 'tiles (5,1) is removed before tile (1,1)' is a problem... But it won't affect path-finding in that it will be considered when a link fails?

alpaca2333: @memononen fair point.

alpaca2333: @memononen hi, i have implemented that in my game and had it tested under dynamicly adding & removing tiles.
Would you approve this work-around? If you would i will migrate that to recast and test it in a few days.

alpaca2333: @memononen hi, would you have a look at this. Now timing problems are gone.
image

alpaca2333: also had it tested under adding & removing tiles 
image
image

alpaca2333: @grahamboree notice that you are assigned a similar issue.

alpaca2333: notice that STL is not approved.

grahamboree:
Hi!  Thanks for this PR.  This is a long-requested enhancement, but unfortunately it's not quite the approach we'd hope for this feature.  We also don't allow the use of STL in Recast for compatibility and portability. For these reasons I'm going to close this PR. I'd be very happy to consider another that addresses the issues Mikko outlined, as well as avoids using STL containers.  Thanks again.

PR #598, opened 2022-12-28
More Recast code cleanup

grahamboree:
Zero functionality changes.  Just formatting changes, local var and parameter name changes, consolidating doc comments and fixing some rule-of-three violations.

Also added a constructor and destructor to rcPolyMeshDetail, since rcAllocPolyMeshDetail and rcFreePolyMeshDetail indicate these can't be trivially constructed or destructed safely like other POD types.

### DIFF
--- a/Recast/Source/Recast.cpp
+++ b/Recast/Source/Recast.cpp
@@ -29,7 +29,6 @@
 namespace
 {
 /// Allocates and constructs an object of the given type, returning a pointer.
-/// TODO: Support constructor args.
 /// @param[in]		allocLifetime	Allocation lifetime hint
 template<typename T>
 T* rcNew(const rcAllocHint allocLifetime)
@@ -74,7 +73,7 @@ void rcContext::log(const rcLogCategory category, const char* format, ...)
 		msg[MSG_SIZE - 1] = '\0';
 
 		const char* errorMessage = "Log message was truncated";
-		doLog(rcLogCategory::RC_LOG_ERROR, errorMessage, (int)strlen(errorMessage));
+		doLog(RC_LOG_ERROR, errorMessage, (int)strlen(errorMessage));
 	}
 	va_end(argList);
 	doLog(category, msg, len);
@@ -88,6 +87,11 @@ void rcContext::doResetLog()
 rcHeightfield* rcAllocHeightfield()
 {
 	return rcNew<rcHeightfield>(RC_ALLOC_PERM);
+}
+
+void rcFreeHeightField(rcHeightfield* heightfield)
+{
+	rcDelete(heightfield);
 }
 
 rcHeightfield::rcHeightfield()
@@ -123,19 +127,14 @@ rcHeightfield::~rcHeightfield()
 	}
 }
 
-void rcFreeHeightField(rcHeightfield* hf)
-{
-	rcDelete(hf);
-}
-
 rcCompactHeightfield* rcAllocCompactHeightfield()
 {
 	return rcNew<rcCompactHeightfield>(RC_ALLOC_PERM);
 }
 
-void rcFreeCompactHeightfield(rcCompactHeightfield* chf)
+void rcFreeCompactHeightfield(rcCompactHeightfield* compactHeightfield)
 {
-	rcDelete(chf);
+	rcDelete(compactHeightfield);
 }
 
 rcCompactHeightfield::rcCompactHeightfield()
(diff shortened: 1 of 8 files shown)


Recast/Source/RecastFilter.cpp
```diff
@@ -16,186 +16,168 @@
...
 	
 	// Mark border spans.
-	for (int y = 0; y < h; ++y)
+	for (int z = 0; z < zSize; ++z)
 	{
-		for (int x = 0; x < w; ++x)
+		for (int x = 0; x < xSize; ++x)
 		{
-			for (rcSpan* s = solid.spans[x + y*w]; s; s = s->next)
+			for (rcSpan* span = heightfield.spans[x + z * xSize]; span; span = span->next)
```
memononen, line 78: In some Recast derivatives I used separate variable for `zStride`, since `z * xSize` registered as bug to me too often :) It does get a bit more verbose, though.

grahamboree: OH nice yeah that's much nicer to read :)

Recast/Source/Recast.cpp
```diff
@@ -262,319 +260,284 @@ rcPolyMesh::~rcPolyMesh()
...
+rcPolyMeshDetail::~rcPolyMeshDetail()
 {
-	if (!dmesh) return;
-	rcFree(dmesh->meshes);
-	rcFree(dmesh->verts);
-	rcFree(dmesh->tris);
-	rcFree(dmesh);
+	rcFree(meshes);
+	rcFree(verts);
+	rcFree(tris);
```
elsid, line 285: This change breaks compatibility with code using older versions of RecastNavigation causing double free since previously there was no destructor and it was required to call `rcFree` for this fields from client code.

grahamboree: I don't quite follow why this would cause a double-free.  This is just moving the calls to `rcFree` the poly mesh members from `rcFreePolyMeshDetail` to `~rcPolyMeshDetail`.  

It sounds like you might not be calling `rcFreePolyMeshDetail` and are instead making the same calls to `rcFree` yourself?  In that case, I don't think it's going to result in a double-free, since you'd be calling `rcFree` on the `rcPolyMeshDetail` instance, which bypasses the destructor.  

The only case I can imagine where this might cause a double-free is if you're calling the destructor explicitly _and_ are calling `rcFreePolyMeshDetail`.

Can you post some usage code that is now a double-free as a result of this change?

Regardless, any double-free issues here could be resolved by zero-ing these pointers after free, which is probably a good idea anyway.

elsid: The problem is that when `rcPolyMeshDetail` is constructed without use of `rcAllocPolyMeshDetail`. It's not used because there is no need to allocate an object of 3 number and 3 pointers on a heap. And in general it's not a good idea to force users to allocate objects on a heap from the performance point of view.

The client code is open source. Basically a combination of [this](https://github.com/OpenMW/openmw/blob/3f6b33820a4b660285e4dc7d83f48e83a84d0a46/components/detournavigator/preparednavmeshdata.hpp#L18) and [this](https://github.com/OpenMW/openmw/blob/3f6b33820a4b660285e4dc7d83f48e83a84d0a46/components/detournavigator/preparednavmeshdata.cpp#L40) triggered the double free. [This](https://gitlab.com/OpenMW/openmw/-/merge_requests/2564) is how I solved the problem to make the code compatible with both versions of the recastnavigation.

> Regardless, any double-free issues here could be resolved by zero-ing these pointers after free, which is probably a good idea anyway.

I don't agree. Double free should not happen in the first place. If it happens there is a flaw in the code logic. Better to fix the flaw rather than hide it by zeroing pointer. Unfortunatelly in my case it's not possible because I need to keep compatibility with multiple recastnavigation versions.

grahamboree: Thanks for the clarification.  

To recap your situation, it comes down to creating a `rcPolyMeshDetail` without calling `rcAllocPolyMeshDetail`, not calling `rcFreePolyMeshDetail`, and manually freeing the three array pointers in `rcPolyMeshDetail` in some custom routine.

To be clear, I'm not necessarily advocating that _just_ null-ing a freed pointer is a good enough solution.  I agree with you that double-frees should be avoided.  My point was more that zero-ing free'd pointers is just good code hygiene and something we should consider.

Moving these `rcFree` calls into the destructor of `rcPolyMeshDetail` makes your use case easier to implement and less prone to memory bugs.  To your point, we shouldn't force people to use the heap, and moving the `rcFree` calls to the destructor means you're not forced to implement them yourself if you're not using `rcFreePolyMeshDetail`, like you've done here.

For the sake of semver and maintaining perfect backwards compatibility, I think I'll move this specific change to the v2.0 branch and keep the old functionality in main.

PR #601, opened 2023-01-25
Fix for out of bounds read in tile cache builder.

AlexanderPolyakov:
Fix for reading data out of bounds in tile cache builder in removeVertex.
While removing a vertex data is read from i+1 index from mesh.verts, which becomes out of bound for the last vertex, which we shouldn't copy over in this case.


mendsley (approved):

grahamboree:
LGTM, thanks for finding this!

PR #611, opened 2023-03-19
Make installing PDBs optional.

### LINKED ISSUE #610: Missing PDBs cause installing to fail.
In some CMake configurations, PDBs may not be generated. For example, by passing the /Z7 flag, debug symbols are embedded in the resulting binaries as opposed to generating a PDB file. This causes the install to fail, which can cause problems for other tools.

andriyDev:
PDBs are not necessarily generated, even in Debug configuration. For example, if the CXX_FLAGS are set to /Z7, debug symbols are embedded, and so the PDBs are not generated. This prevents installing from failing by marking these PDBs as optional.

Fix #610


grahamboree:
Much appreciated, thanks!

PR #617, opened 2023-03-25
fixed a typo (countour -> contour)

FcAYH:
just a typo.


grahamboree:
Thanks for catching this!

PR #619, opened 2023-03-30
fix invalid assert dtAssert(npath < m_maxPath)

nicknide:
In the function dtPathCorridor::setCorridor
```
	dtAssert(npath < m_maxPath);
```
the assertion failure caused a crash on our server.

After analyzing the source code of this file, it was found that the value assigned to m_npath was used as a size rather than an index, and all the places where it was used would not lead to direct access.

in function dtCrowd::updateMoveRequest
```
    // Make space for the old path.
    if ((npath-1)+nres > m_maxPathResult)
        nres = m_maxPathResult - (npath-1);
    
    memmove(res+npath-1, res, sizeof(dtPolyRef)*nres);
    // Copy old path in the beginning.
    memcpy(res, path, sizeof(dtPolyRef)*(npath-1));
    nres += npath-1;
```
there is a possibility of `nres == m_maxPathResult`, so this value is normal.

If there is any misunderstanding, please correct me. Thank you.

### DIFF
--- a/DetourCrowd/Source/DetourPathCorridor.cpp
+++ b/DetourCrowd/Source/DetourPathCorridor.cpp
@@ -512,7 +512,7 @@ void dtPathCorridor::setCorridor(const float* target, const dtPolyRef* path, con
 {
 	dtAssert(m_path);
 	dtAssert(npath > 0);
-	dtAssert(npath < m_maxPath);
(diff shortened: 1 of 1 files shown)


grahamboree:
Thanks!

Issue #604, opened 2023-03-10
Make RC_MAX_LAYERS  and RC_MAX_NEIS configurable at build time


bmorel:
[Those](https://github.com/recastnavigation/recastnavigation/blob/b43c82c8087f7e277077dcb7cb1e6c626d5550ce/Recast/Source/RecastLayers.cpp#L29-L32) are currently global constants, which require modifying recast's source itself (see [this issue](https://github.com/Unvanquished/Unvanquished/issues/2403) from a game that uses recast).

Would it be ok to make it possible to make this a cmake option instead? This way, it would be possible for users to at least change the value at compilation time without having to modify recast's code directly.

grahamboree:
Hi!  I'm not terribly familiar with cmake's build configuration system, but I assume it's essentially setting C preprocessor variables like other build systems do.

I'd be open to moving these integers to `#define`'s instead since it'd expose configuration to build systems like cmake, assuming that's what cmake options are operating with.

PR #621, opened 2023-04-09
cmake configure for RC_MAX_LAYERS and RC_MAX_NEIS

bmorel:
implements feature mentioned in #604 and discussed in #605

I added RC_MAX_NEIS in because why not. It could be improved by adding actual doc strings, but I'm not sure what I should write in there.
This also makes sure that RC_MAX_LAYERS can not be > 255, because why not too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,17 @@ endif()
 include(GNUInstallDirs)
 
 configure_file(
+    "${RecastNavigation_SOURCE_DIR}/config.h.in"
+    "${RecastNavigation_BINARY_DIR}/config.h")
+include_directories( ${PROJECT_BINARY_DIR} )
+SET(RC_MAX_LAYERS 63 CACHE STRING "Must be 255 or smaller (not 256) because layer IDs are stored as a byte where 255 is a special value.")
+SET(RC_MAX_NEIS 16 CACHE STRING "" FORCE)
+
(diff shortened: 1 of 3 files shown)


grahamboree:
Hi, thanks for this!  Unfortunately this change doesn't work if you're not building with Cmake (see the CI output above).  Is there a version of this that exposes these values to CMake but also lets people not using Cmake set them as well?

bmorel:
So, CI does  uses a different build system? Well, I guess I'll just do the old "ifndef define " then

grahamboree:
Yes, Recast has always used premake as a build system for the demo project.  CMake is a recent addition since some folks were interested in supporting it.  In the games industry, Recast is most often embedded directly as source files into projects sidestepping build systems altogether.  Because of this it's important that Recast builds in other contexts as well as CMake.  The CI jobs run the premake and CMake pipelines for all 3 OSes using GCC, Clang, and MSVC.  We try to ensure that we're supporting the broadest set of tool chains as possible.  You can see more about building recast with tools other than CMake in the project readme.

illwieckz:
Hi @grahamboree!

I submitted an alternative implementation for this feature, in a way it's totally agnostic regarding to the build system, and in a way it's entirely up to the third-party project integrating Recast to set his own build system switches to configure the max values:

- https://github.com/recastnavigation/recastnavigation/pull/624

PR #622, opened 2023-04-10
fix: a offmeshlink should alloc 3 dtLinks at start tile and 1 dtLink at end …

zkz196:
The number of dtLinks assigned to offmeshlink is incorrect, resulting in loss of connections between polygons


grahamboree:
Hi!  Thanks for the PR.  Can you provide some more information about the specific issue this is solving?

zkz196:
image
as you see, the old code calculated maxLinkCount as
edgeCount(4) + portalCount(2)*2 + offMeshConLinkCount(8)*2 = 24.
In fact, offMeshLink will be used
baseOffMeshLinks(8) * 2 + connectExtOffMeshLinks(8) = 24 
so the start polygon do not link the end polygon directly

eugeneko:
@zkz196 Out of curiosity, why do we need 3 links on start and 1 on end? I almost never use offmesh, I want to understand the structure of links better.

PR #624, opened 2023-04-11
RecastLayers: set RC_MAX_LAYERS and RC_MAX_NEIS as optional defines

illwieckz:
It makes possible for a project integrating Recast to set custom values via CXXFLAGS while remaining build system agnostic.

Type checking is kept.

This is an alternative implementation for:

- https://github.com/recastnavigation/recastnavigation/pull/621

Here is an example of how a project integrating Recast may customize `RC_MAX_LAYERS` once this patch is merged. In this example the build-system specific code is entirely externalized, keeping the patch on Recast side small and unlikely to require any maintenance in the future:

- https://github.com/Unvanquished/Unvanquished/pull/2620

This also means that once this PR is merged a third-party project will start relying on this feature and then this feature will not be orphan. This wanted feature fulfills an actual need and merging it upstream will reduce code fragmentation.

### DIFF
--- a/Recast/Source/RecastLayers.cpp
+++ b/Recast/Source/RecastLayers.cpp
@@ -29,8 +29,21 @@
 
 // Must be 255 or smaller (not 256) because layer IDs are stored as
 // a byte where 255 is a special value.
-static const int RC_MAX_LAYERS = 63;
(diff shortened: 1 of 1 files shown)


grahamboree:
Thanks for this!  Looks great.

PR #625, opened 2023-04-16
Mark input data structures as const.

andriyDev:
Marks some functions which use inputs as const, but that are not marked const.

These functions do not modify their inputs so they should be marked as const to avoid confusion about ownership.

This change does technically modify the public API, but for C++ code at least no uses should change.


grahamboree:
LGTM Thanks!

PR #626, opened 2023-04-28
Fix spans being filtered even if they have just enough clearance.

andriyDev:
The function comment mentions that heights LESS THAN the walkable height will be marked unwalkable. The previous implementation filtered out heights with exactly walkable height space.


grahamboree:
LGTM thanks for this!

PR #627, opened 2023-04-29
Add a CMake option to enable 64bit dtPolyRefs for Detour.

andriyDev:
Previously, setting this option required either enabling this option by modifying the source, or doing some CMake hacking to inject the variables. Now this is a convenient option.


grahamboree:
Thanks!  LGTM.  Would be good to do something similar for `DT_VIRTUAL_QUERYFILTER`

PR #635, opened 2023-06-18
Fix out of bounds access in dtMergeCorridorStartMoved and add tests

elsid:
`size` can become negative if `req > maxPath`. This may happen when `visited` buffer is larger than `path` buffer.

Add tests to cover different use cases of the function including `Should add visited points not present in path up to the path capacity` to cover the fix.

List tests files explicitly. When new file is added CMake does not add it to the already generated list if GLOB is used.

Update catch to `v3.3.2` to get `RangeEquals` matcher making code for matching arrays simpler.


grahamboree:
Thanks for this fix and for the unit tests as well!  Unfortunately I think there's something about the catch upgrade that's causing the windows unit tests job to stall out.  Is there any chance you could split out the catch upgrade into a separate PR?

elsid:
Added a fix for out of bounds access and now all tests are passing.

grahamboree:
Thanks!  Love that more of the code is getting unit tests like this. 🎉

PR #636, opened 2023-06-18
Improved variable naming in RecastArea.cpp

grahamboree:
This should be a non-functional change apart from one bug I found in `rcOffsetPoly` where [a call to normalize a vector](https://github.com/recastnavigation/recastnavigation/blob/6dc1667f580357e8a2154c28b7867bea7e8ad3a7/Recast/Source/RecastArea.cpp#L489) was dividing it by the squared magnitude rather than the magnitude.

### DIFF
--- a/Recast/Source/RecastArea.cpp
+++ b/Recast/Source/RecastArea.cpp
@@ -513,79 +513,113 @@ void rcMarkConvexPolyArea(rcContext* context, const float* verts, const int numV
 
 int rcOffsetPoly(const float* verts, const int numVerts, const float offset, float* outVerts, const int maxOutVerts)
 {
-	// TODO (graham): Better variable names here.
 	const float MITER_LIMIT = 1.20f;
+    const float EPSILON = 1e-6f;
 
-	int n = 0;
+	int numOutVerts = 0;
 
-	for (int i = 0; i < numVerts; i++)
+	for (int vertIndex = 0; vertIndex < numVerts; vertIndex++)
 	{
-		const int a = (i + numVerts - 1) % numVerts;
-		const int b = i;
-		const int c = (i + 1) % numVerts;
-		const float* va = &verts[a * 3];
-		const float* vb = &verts[b * 3];
-		const float* vc = &verts[c * 3];
-		float dx0 = vb[0] - va[0];
-		float dy0 = vb[2] - va[2];
-		float d0 = dx0 * dx0 + dy0 * dy0;
-		if (d0 > 1e-6f)
+        // Grab three vertices of the polygon.
+		const int vertIndexA = (vertIndex + numVerts - 1) % numVerts;
+		const int vertIndexB = vertIndex;
+		const int vertIndexC = (vertIndex + 1) % numVerts;
+		const float* vertA = &verts[vertIndexA * 3];
+		const float* vertB = &verts[vertIndexB * 3];
+		const float* vertC = &verts[vertIndexC * 3];
+
+        // From A to B on the x/z plane
+		float ABdeltaX = vertB[0] - vertA[0];
+		float ABdeltaZ = vertB[2] - vertA[2];
+		float ABsqMag = ABdeltaX * ABdeltaX + ABdeltaZ * ABdeltaZ;
+		if (ABsqMag > EPSILON) // Is the squared magnitude of the distance between A and B greater than zero?
 		{
-			d0 = 1.0f / rcSqrt(d0);
-			dx0 *= d0;
-			dy0 *= d0;
-		}
-		float dx1 = vc[0] - vb[0];
-		float dy1 = vc[2] - vb[2];
-		float d1 = dx1 * dx1 + dy1 * dy1;
-		if (d1 > 1e-6f)
-		{
-			d1 = 1.0f / rcSqrt(d1);
-			dx1 *= d1;
-			dy1 *= d1;
-		}
-		const float dlx0 = -dy0;
-		const float dly0 = dx0;
-		const float dlx1 = -dy1;
-		const float dly1 = dx1;
-		float cross = dx1 * dy0 - dx0 * dy1;
-		float dmx = (dlx0 + dlx1) * 0.5f;
-		float dmy = (dly0 + dly1) * 0.5f;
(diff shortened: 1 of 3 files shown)


Recast/Source/RecastArea.cpp
```diff
@@ -513,79 +513,113 @@ void rcMarkConvexPolyArea(rcContext* context, const float* verts, const int numV
...
+        // that the result should be beveled.
+		bool bevel = BNormalSqMag * MITER_LIMIT * MITER_LIMIT < 1.0f;
+
+        // Now we normalize the BNormal vector
+		if (BNormalSqMag > EPSILON)
 		{
-			const float scale = 1.0f / dmr2;
-			dmx *= scale;
-			dmy *= scale;
+			const float inverseMag = 1.0f / rcSqrt(BNormalSqMag);
```
memononen, line 582: This is not correct, and it's not normal, it is "middle" or "average" offset vector.

The the middle vector is calculated as if the _edges_ were offset and we would calculate intersection of neighbour edges at each vertex. That is, the sharper the corner, the further the offset is projected (thus the bevel limit too, or else it would shoot to sunset). If we normalize the offset, then the corners will get squashed.

Recast/Source/RecastArea.cpp
```diff
@@ -513,79 +513,113 @@ void rcMarkConvexPolyArea(rcContext* context, const float* verts, const int numV
...
+        // Grab three vertices of the polygon.
+		const int vertIndexA = (vertIndex + numVerts - 1) % numVerts;
+		const int vertIndexB = vertIndex;
+		const int vertIndexC = (vertIndex + 1) % numVerts;
+		const float* vertA = &verts[vertIndexA * 3];
+		const float* vertB = &verts[vertIndexB * 3];
+		const float* vertC = &verts[vertIndexC * 3];
+
+        // From A to B on the x/z plane
+		float ABdeltaX = vertB[0] - vertA[0];
```
memononen, line 533: I think these names does not make things any more clear, some suggestions:
ABdeltaX/Y -> prevSegmentDirX/Y
BCdeltaX/Y -> curSegmentDirX/Y

ABperpX/Y -> prevSegmentNormX/Y
BCperpX/Y -> curSegmentNormX/Y

memononen: Maybe add a helper function for the normalization, it's known concept but makes a lot of noise in this function.

Recast/Source/RecastArea.cpp
memononen: You can quickly verify that on paper with a 90° corner. preSegNorm=(-1,0), curSegNorm(0,1), should give you (-1,1), while normalized version would be (-0.707, 0.707).

Miter limit should be the same as commonly used in graphics (e.g. https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit). If my math is right, 1.2 is about 17.5° corner.

memononen: I tried to find source for the algo, but failed. I picked it up from some code long ago (could have been libart). Since the PostScript line miter limit comes so naturally out of it, maybe it comes from that lineage.

One "derivation" I did for myself at some point was that if you have segment normals `na` and `nb`, you calculate the bisector `na+nb`, and then project the bisector to `na`, you get the bisectors length `d` on `na`. Finally you find a scaling factor `1/d` so that the projected bisector length is 1 on `na`.

grahamboree: Aaaaah this makes a ton of sense.  Thanks for clarifying!

Recast/Source/RecastArea.cpp
```diff
@@ -16,573 +16,700 @@
...
-			outVerts[n*3+0] = vb[0] - dmx*offset;
-			outVerts[n*3+1] = vb[1];
-			outVerts[n*3+2] = vb[2] - dmy*offset;
-			n++;
+			}
+
+            // Move B along the miter direction by the specified offset.
+			outVerts[numOutVerts * 3 + 0] = vertB[0] - cornerMiterX * offset;
+			outVerts[numOutVerts * 3 + 1] = vertB[1];
+			outVerts[numOutVerts * 3 + 2] = vertB[2] - cornerMiterZ * offset;
```
grahamboree, line 625: @memononen Digging more into this, I'm still a bit confused why this is _subtracting_ the scaled miter offset from vertB?  Seems like that implies either positive values for `offset` should _shrink_ the polygon instead of expand it, _or_ the bisector here should be actually pointing _into_ the polygon?

e.g.
(I know the code would bevel a 90 degree corner like this.  Just using these numbers to make the math easy.)

```
            C
B *---------*
  |
  |
  |
  |
A *


A = <0, 0, 0>
B = <0, 0, 10>
C = <10, 0, 10>

// Normalized segment directions
prevSegmentDir = normalize(B - A) = <0, 1>
currSegmentDir = normalize(C - B) = <1, 0>

// Segment normals pointing out of the poly (ccw perp vector assuming cw polygon winding order)
prevSegmentNorm = <-1, 0>
currSegmentNorm = <0, 1>

cornerMiter = (prevSegmentNorm + currSegmentNorm) / 2 = <-.5, .5>
cornerMiterSqMag = .5

// The final scaled miter vector (bisector) points out of the polygon
cornerMiter = cornerMiter / cornerMiterSqMag = <-1, 1>

// given an offset value of 2, B is offset correctly but in the wrong direction?
outVert = B - cornerMiter * offset = <0, 0, 10> - <-1, 0, 1> * 2 = <2, 0, 8>

// I'd expect the final result here to be <-2, 0, 12> instead
```
I suspect I'm missing something critical here, or maybe I'm getting turned around with coordinate system handedness?

grahamboree: The above is kinda confusing.... when I use 2-component vectors `<1, 2>` it's really an XZ plane vector so think of it as `<1, 0, 2>`

memononen: The code is the same as in https://github.com/recastnavigation/recastnavigation/blob/main/RecastDemo/Source/imguiRenderGL.cpp#L61 (which is recycled from another rendering lib, which eventually turned into nanovg),
but adjusted to what ever winding the polygon happened to have.

PR #637, opened 2023-06-19
Removed fixupCorridor and replaced with dtMergeCorridorStartMoved

### LINKED ISSUE #634: RecastDemo should use dtMergeCorridorStartMoved rather than re-implementing it
From #633 

The `fixupCorridor` function in Recast Demo is simply a copy of `dtMergeCorridorStartMoved`.  Since the latter is part of the recast crowd API, we should use that in the demo instead.

jozamudi:
In the function dtMergeCorridorStartMoved is re-implemented in NavMeshTesterTool.cpp as fixupCorridor. 

I removed fixupCorridor and replaced with dtMergeCorridorStartMoved.

Fixes #634


grahamboree:
Looks great.  Thanks for this!

Issue #586, opened 2022-11-26
Impromptu fix for off-mesh connections with long distance


niji-k:
![image](https://user-images.githubusercontent.com/119095278/204074851-5bb7e1a5-54d2-4d8c-91aa-e4155ee15412.png)

I'm a junior developer, using Recast Navigation for my group's game.
I was told that these off-mesh connections won't work with long distances between endpoints, and that the level designer must adjust the tile size manually to make it work.

After some googling and reading through lines of code, I realized that these connections would only work if they are within a single tile, or connecting two adjacent tiles. I presume this is probably to block issues that could emerge when trying to add connection(s) before the two tiles containing two endpoints of said connection are fully loaded.

I was able to produce a working fix, by forcing the code to add connection after all tiles are fully loaded.
This fix is impromptu as the title suggests, and if my testing is right, it does not guarantee that the path is optimal.
Please take a look at this commit made to a fork, and tell me what you think about it: https://github.com/niji-k/recastnavigation/commit/13eb591b1b752de2e5a78137580b7aeec106eeab

Below are screenshots made while testing, showing some possible cases with modified behavior from this fix.
![image](https://user-images.githubusercontent.com/119095278/204075276-9ce5160c-fb8a-4eff-9ad2-f1535e692131.png)
(Pathfinding successfully utilized the off-mesh connection given.)

![image](https://user-images.githubusercontent.com/119095278/204075383-98807eea-acd7-4f0c-8861-bc67f9536c73.png)
(Suboptimal path: I have no idea why the connection isn't used.)

![image](https://user-images.githubusercontent.com/119095278/204075620-d74914ab-cbd5-4f40-83ab-8b92266499f5.png)
![image](https://user-images.githubusercontent.com/119095278/204075626-d394da38-b15a-4a7e-92c3-64bd53bc1799.png)
(One-way connections can also be used: notice how swapping the start/end pos changes the path.)

P.S. I'm definitely new to this Github stuff, let alone Git, so any advices regarding the Github workflow is highly appreciated.

grahamboree:
Thanks for this!  Could you instead submit this as a pull request?  There's some documentation [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) on how to do that in GitHub.

Issue #195, opened 2016-03-16
Are OffMeshLinks Limited to Tile Neighbours?


graehu:
After having a read through of dtNavMesh::addTile() it looks like it's not possible to connect to tiles that aren't [neighbours][1]. If that is the case, what would the best way to implement a teleporter be?

[1]:https://github.com/recastnavigation/recastnavigation/blob/master/Detour/Source/DetourNavMesh.cpp#L967

Janiels:
Yes, off mesh connections are only supported within tiles and to adjacent tiles. The best way to implement a teleporter would probably be to fix this, or to use some hierarchical path planning on top of Detour.

One possible solution would be to expand the code in `addTile` to connect off mesh links to/from all other tiles in the mesh, although this might be a bit slow but maybe it's acceptable for your use case.
Another solution is to fully implement dynamic links. #129 is kind of related - @memononen mentions using a pool for links, which could support dynamic off mesh connections, including connections farther than just an adjacent tile.

Supporting dynamic links and by extension non-adjacent off mesh connections is a feature we definitely want to have at some point, but haven't gotten around to implementing. PR's are always welcome!

graehu:
Thanks for getting back to me. I'm using detour through UnrealEngine 4, so any changes you make wouldn't make it in for quite some time probably. For now, I'm going to increase my tile size to accommodate for the distance of teleporters in each of my levels. Thankfully they're quite small.

Though, if I think about changing how links in my version of UE4 work I'll let you know. Based on the link exit coordinates I think it could be relatively trivial to find the exit tile rather than simply checking neighbours. But maintaining state whenever that tile is modified may be less trivial, my understanding of Detour isn't good enough to fully gauge the problem at this point sadly. :P

I'll keep an eye out for cool changes to Detour either way though because it's awesome!

Janiels:
Detour only stores off mesh connection in the starting tile. This means that if you have an off mesh connection pointing from tile A to B and you have loaded tile A first it will not be able to connect the off mesh connection when tile B is loaded. This can potentially give you some inconsistent results depending on the tile load order, should you choose to go with the "find exit tile" solution.
A potential fix to that problem could be to store off mesh connections in both tiles.

Issue #477, opened 2021-04-05
Off-mesh connections not working when ends are far apart


TheSHEEEP:
See the following image:
![pathfinding_problem](https://user-images.githubusercontent.com/1766205/113543688-92f23000-95ef-11eb-9b48-2c01a67fa826.png)

It is very clear that off-mesh connections are only working when the ends are close together.  
I'm certain there is some kind of logic to this behavior.

In the code, I can see that ends are drawn red (i.e. not working) according to this logic:
```
// Check to see if start and end end-points have links.
bool startSet = false;
bool endSet = false;
for (unsigned int k = p->firstLink; k != DT_NULL_LINK; k = tile->links[k].next)
{
	if (tile->links[k].edge == 0)
		startSet = true;
	if (tile->links[k].edge == 1)
		endSet = true;
}
```
How these "links" are set when the tile mesh is built, though, I fail to understand, as that is a bit too in-depth in the code for me to get behind at this moment.

What's more important is that I don't see a design reason for this restriction.  
Why not make those connections work from any point to any other point? Why limit the range?

Any explanation or idea to get around this restriction would be very helpful.

cybroxtu:
According to the offmesh-link initialization code in `dtNavMesh::addTile`, only those offmesh-links connecting neighbour tiles are valid.

https://github.com/recastnavigation/recastnavigation/blob/c5cbd53024c8a9d8d097a4371215e3342d2fdc87/Detour/Source/DetourNavMesh.cpp#L1038-L1049

TheSHEEEP:
That seems like a very arbitrary and unnecessary restriction, unless I'm not seeing something here.

jakobbotsch:
I agree, this would be a nice restriction to lift. We need to maintain a table of unlinked offmesh connections to do so to ensure that loading tiles in either order works (that's probably the main reason the restriction is there currently). #195 is related.

bmorel:
Yes, this would be very useful, as it would enable bots to travel through teleporters or to use jumppads (that sends them far away).

PR #645, opened 2023-07-05
Support for long range off-mesh connection support

niji-k:
This lifts the restriction for off-mesh connections for tiled navigation meshes
where the two endpoints for a connection must be located in the same tile, or in adjacent tiles.
This is done by connection the off-mesh links AFTER all tiles have been built.
You may find more detail on: https://github.com/recastnavigation/recastnavigation/issues/195#issuecomment-198132561
> Detour only stores off mesh connection in the starting tile. This means that if you have an off mesh connection pointing from tile A to B and you have loaded tile A first it will not be able to connect the off mesh connection when tile B is loaded. This can potentially give you some inconsistent results depending on the tile load order, should you choose to go with the "find exit tile" solution.
> A potential fix to that problem could be to store off mesh connections in both tiles.

This "fix" is only crudely worked on as this issue suggests; an impropmtu fix.
https://github.com/recastnavigation/recastnavigation/issues/586#issue-1465066364

More specifically, the change was only made to serve as a showcase, and although this should be working as intended with *building the whole tile mesh*, the same may not be applied to:

- Building the tile mesh *partially*, i.e. building only one tile at a time,
- & loading the tile mesh from a binary file.

To make it work for cases above, further exploration on this method of approach should be necessary.


niji-k:
That should be easy to overcome, one simply would need to create a list of all unlinked off-mesh connections, as said in the comment below:
https://github.com/recastnavigation/recastnavigation/issues/477#issuecomment-851047403

The only reason for my commit disabling `connectExtOffMeshLinks()` at said position, is that the behavior of two functions `connectExtOffMeshLinks()` and `connectGlobalOffMeshLinks()` are almost identical to each other.
As for the benefits, I'm not so sure. For all I know, it's just a matter between linking the tiles during the tile loading pass, or after said pass.

RGreenlees:
This is excellent, are you planning an implementation that supports the tile cache (Sample_TempObstacles) as well? I'm not sure how to integrate this into the tile cache.

mastahg:
Some feedback on this.

1)It works

2)The documentation for the function is incorrect in my understanding, target is actually where the starting point is, and tile is the end point. It's reverse what you would expect.

3)The sample connection establishment is extremely slow for meshes with large amounts of tiles

First I moved it to be a member function of dtNavMesh, as it's something I think should be supplied and simply called by anyone using the library.

I then focused on optimizing the function call.

I altered and rearranged the checks, continuing the loop if
 ```
if (target->header->offMeshConCount == 0)
 continue
```
this was the first major speedup as there's no point looping thru the X range if there's no offmesh connections in the target tile.

I then added a new variable to the dtMeshTile class unconnectedOffMeshCount, and during tile construction set this to the header->offMeshConCount.
This value would be decremented inside the end of connectGlobalOffMeshLinks and connectExtOffMeshLinks.

This value is also checked, at the end of dtAddtile and if it has a value > 0 then m_hasOffMesh is set to true.

This gives a huge performance benefit for meshes where there are only connections inside the same tile that we don't have to scan the entire tile range to find start/end connections.

Further possible improvements:

It maybe worthwhile to keep the neighbor checks in, as searching neighbors is faster then scanning the entire tile range.

It maybe worthwhile to import the rcVector class and add the tiles where unconnectedOffMeshCount > 0 that way we don't scan the entire tile range

It maybe faster to use a navmeshquery to find the what the end tile would be for a given off mesh connection, rather then scanning all the given tiles.

```
void dtNavMesh::ConnectGlobalOffMeshLinks()
{
	if (m_hasOffMesh)
	{
		const int tileNum = this->getMaxTiles();
		for (int y = 0; y < tileNum; ++y)
		{
			const auto target = this->getTile(y);
	
			if (!target->data)
				continue;
	
			if (target->unconnectedOffMeshCount == 0)
				continue;
	
			for (int x = 0; x < tileNum; ++x)
			{
				if (x == y) 
					continue;

				const auto tile = this->getTile(x);
				
				if (!tile->data) 
					continue;

				this->connectGlobalOffMeshLinks(tile, target);

				if (target->unconnectedOffMeshCount == 0)
					break;

			}
		}
	}
}
```

**Comment by mastahg** [comment]
So I played around with this a bit more

I ported over rcVectorBase and renamed to dtVectorBase etc, and during the dtnavmesh::addtile, checked if header->offMeshConCount > 0 added it to the vector.

Then a simple function

```

void dtNavMesh::ConnectGlobalOffMeshLinks()
{
	if (m_hasOffMesh)
	{
		for(const auto target : tilesWithOffmeshConnections)
		{
			GlobalOffMeshLinks(target);
		}

}

```
 and

smashed together the logic from dtNavMeshQuery::queryPolygons and the previous function create the following:

```
void dtNavMesh::GlobalOffMeshLinks(dtMeshTile* target)
{
	if (!target) 
		return;

	for (int i = 0; i < target->header->offMeshConCount; ++i)
	{
		dtOffMeshConnection* targetCon = &target->offMeshCons[i];

		dtPoly* targetPoly = &target->polys[targetCon->poly];
		// Skip off-mesh connections which start location could not be connected at all.
		if (targetPoly->firstLink == DT_NULL_LINK)
			continue;

		const float ext[3] = { targetCon->rad, target->header->walkableClimb, targetCon->rad };


		float bmin[3], bmax[3];
		dtVsub(bmin, &targetCon->pos[3], ext);
		dtVadd(bmax, &targetCon->pos[3], ext);

		// Find tiles the query touches.
		int minx, miny, maxx, maxy;
		this->calcTileLoc(bmin, &minx, &miny);
		this->calcTileLoc(bmax, &maxx, &maxy);

		static const int MAX_NEIS = 32;
		dtMeshTile* neis[MAX_NEIS];

		for (int y = miny; y <= maxy; ++y)
		{
			for (int x = minx; x <= maxx; ++x)
			{
				const int nneis = this->getTilesAt(x, y, neis, MAX_NEIS);
				{
					auto tile = neis[j];
					// Find polygon to connect to.
					const float* p = &targetCon->pos[3];
					float nearestPt[3];
					dtPolyRef ref = findNearestPolyInTile(tile, p, ext, nearestPt);
					if (!ref)
						continue;
					// findNearestPoly may return too optimistic results, further check to make sure. 
					if (dtSqr(nearestPt[0] - p[0]) + dtSqr(nearestPt[2] - p[2]) > dtSqr(targetCon->rad))
						continue;
					// Make sure the location is on current mesh.
					float* v = &target->verts[targetPoly->verts[1] * 3];
					dtVcopy(v, nearestPt);

					// Link off-mesh connection to target poly.
					unsigned int idx = allocLink(target);
					if (idx != DT_NULL_LINK)
					{
						dtLink* link = &target->links[idx];
						link->ref = ref;
						link->edge = (unsigned char)1;
						link->side = 0xff;
						link->bmin = link->bmax = 0;
						// Add to linked list.
						link->next = targetPoly->firstLink;
						targetPoly->firstLink = idx;
					}

					// Link target poly to off-mesh connection.
					if (targetCon->flags & DT_OFFMESH_CON_BIDIR)
					{
						unsigned int tidx = allocLink(tile);
						if (tidx != DT_NULL_LINK)
						{
							const unsigned short landPolyIdx = (unsigned short)decodePolyIdPoly(ref);
							dtPoly* landPoly = &tile->polys[landPolyIdx];
							dtLink* link = &tile->links[tidx];
							link->ref = getPolyRefBase(target) | (dtPolyRef)(targetCon->poly);
							link->edge = 0xff;
							link->side = (unsigned char)(0xff);
							link->bmin = link->bmax = 0;
							// Add to linked list.
							link->next = landPoly->firstLink;
							landPoly->firstLink = tidx;
						}
					}
				}
			}
		}
	}
}
```
this new method is exceptionally fast, two orders of magnitude faster for meshes with large tile counts, and handles all connections just fine so I even went and disabled the self tile check inside addtile, so that all offmesh connections are handled thru this one function.

Concerns that are outside my usecase:
Removing tiles and then readding.

niko-dellic:
bump, this would be great to have

EvanBalster:
+1.  This would be lovely to have for my non-euclidean game (portal rendering all the way down).

Looks like [dtNavMesh::connectGlobalOffMeshLinks](https://github.com/recastnavigation/recastnavigation/pull/645/changes#diff-a5066bef3adf7c8c5033c79c2807275fc0cd41686c01fe81f5a736a48448d535R1058) does most of the lifting here, and it's essentially just an extra method in `dtNavMesh`.  Interesting, interesting.

EvanBalster:
I concocted a drop-in solution for projects that need this functionality but don't want to work with a fork of the library.  It's based on this PR and works by injecting a single private method into `dtNavMesh`.

See here:  https://gist.github.com/EvanBalster/ea51ddf9c80c3caf730ef83a95a638e2

PR #648, opened 2023-07-20
Update catch to v3.4.0

elsid:
To get `RangeEquals` matcher making code for matching arrays simpler.

* Files are copied from https://github.com/catchorg/Catch2/releases/tag/v3.4.0.
* `catch_amalgamated.hpp` is renamed to `catch_all.hpp` to match existing file.


grahamboree:
Thanks!

PR #661, opened 2023-09-02
Minor refactor in RecastMesh.cpp

Gamemechanicwow:
There is no need for ELSE after an IF statement that always ends with a return.


grahamboree:
Thanks!

Issue #669, opened 2023-09-29
[Detour] Incorrect layout of tile links in DT_POLYREF64 mode


m-bronnikov:
## What?

Addition of tile to navmesh sometimes works incorrect In *Detour* built in `DT_POLYREF64` mode: 

After call of `addTile` method in `dtNavMesh`, pointer to array of `links` in resulting `dtMeshTile` may has incorrect alignment. In cpp, if pointer has incorrect alignment, access to data is **undefined behaviour**.

To reproduce that issue you may:
0. Build RecastDemo with `-DRECASTNAVIGATION_DT_POLYREF64=ON` option
1. Add breakpoint or dbg check after following line:
https://github.com/recastnavigation/recastnavigation/blob/cd898904b72a300011fbb24d578620bafa08ef2c/RecastDemo/Source/Sample_TileMesh.cpp#L761
2. Launch `RecastDemo`, choose `TileMesh` sample with any navmesh and click `Build`
3. Check `links` pointer value (or throw some error if value incorrect).

My environment is x64-bit Ubuntu 18.04 LTS, but bug also checked at Windows 10.

## Why?

The reason why this bug happens in following lines:
https://github.com/recastnavigation/recastnavigation/blob/cd898904b72a300011fbb24d578620bafa08ef2c/Detour/Source/DetourNavMesh.cpp#L978-L996

All sizes here alinged by 4 byte, but some tipes (i.e. `dtLink`) have `alignof` equal to 8 bytes if `DT_POLYREF64` is `ON`.

So, `dtGetThenAdvanceBufferPointer` provides pointer with incorrect alignment to `links` pointer.

m-bronnikov:
I did a pull request which should fix that behaviour and compute pointers values correctly.

PTAL: #670

grahamboree:
Thank you for the detailed writeup!  I unfortunately haven't been able to follow your repro steps as I'm not sure what you're looking for in step 4.  I appreciate its alignment related, but it's unclear how this problem manifests.  

Could you provide some more specifics in your repro steps, especially around why you're setting the breakpoint where you are, and what exactly you're looking at that's incorrect?

m-bronnikov:
Hi @grahamboree, thank you to your reply.

I will try to be more clear this time in my repro steps and especially what's wrong in it.

1. Please take a look at my local commit https://github.com/m-bronnikov/recastnavigation/commit/ba5c07205e7ad7d71dd0ef66b06c207a3852c815

[...]

PR #670, opened 2023-09-29
[Detour] Compute tile pointers aligns correctly

m-bronnikov:
This commit fixes incorrect align computation for data in NavMesh's tile. Commit not changes user API.

Fixes: #669

### DIFF
--- a/Detour/Include/DetourCommon.h
+++ b/Detour/Include/DetourCommon.h
@@ -462,6 +462,13 @@ inline unsigned int dtIlog2(unsigned int v)
 
 inline int dtAlign4(int x) { return (x+3) & ~3; }
 
+template <typename TypeToBeAlignedFor>
+int dtAlignForType(int x) 
+{ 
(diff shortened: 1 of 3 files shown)


grahamboree:
Thanks for this.  While I agree with what you're trying to do here I think this needs some work before it can be accepted.  There's a number of compilation issues on non-Windows platforms as well because clang and gcc throw errors when passing template parameters to `alignof`.

m-bronnikov:
> Thanks for this. While I agree with what you're trying to do here I think this needs some work before it can be accepted. There's a number of compilation issues on non-Windows platforms as well because clang and gcc throw errors when passing template parameters to `alignof`.

The reason of fail is building by c++98 standart, it was a litle bit unexpected =)

I agree with you that I need to make some additional work on this PR. After fixes I will create new PR if you don't mind.

grahamboree:
Awesome, thanks!

PR #672, opened 2023-10-29
Code cleanup and small optimizations in RecastFilter.cpp rcFilterLedgeSpans

zz2108828:
Because span.smax is always > 0, bot > 0 as well, and (-walkableClimb - bot) is always < -walkableClimb. Furthermore, as long as minNeighborHeight < -walkableClimb' at least once, there is no need to continue the traversal.


Recast/Source/RecastFilter.cpp
```diff
@@ -96,42 +96,47 @@ void rcFilterLedgeSpans(rcContext* context, const int walkableHeight, const int
...
+							minNeighborHeight = rcMin(minNeighborHeight, accessibleNeighbourHeight);
 
 							// Find min/max accessible neighbour height. 
-							if (rcAbs(neighborBot - bot) <= walkableClimb)
+							if (rcAbs(accessibleNeighbourHeight) <= walkableClimb)
 							{
 								if (neighborBot < accessibleNeighborMinHeight) accessibleNeighborMinHeight = neighborBot;
 								if (neighborBot > accessibleNeighborMaxHeight) accessibleNeighborMaxHeight = neighborBot;
 							}
+							else if (accessibleNeighbourHeight < -walkableClimb)
```
grahamboree, line 136: Is it safe to break out of the loop here? `minNeighborHeight` might not get set correctly if there are multiple adjacent walkable areas that overlap more than `walkableClimb` with the walkable area above the current span.

grahamboree: oh nvm I guess since iteration order is bottom to top it's fine

grahamboree (approved):

grahamboree:
Ty!

elsid:
This PR changed the result of navmesh generation. Does not look like it broke something but it's unclear if this was intentional. If this is an optimization of the generation process it should not affect the data. One more reason to do https://github.com/recastnavigation/recastnavigation/issues/678 .

What exactly has been changed in the navmesh data requires additional investigation but the change is visible because different paths are generated for the same input (to recast to generate navmesh to be used by detour). In a test case where I've noticed this just one point is added in the middle of the path and it looks legitimate.

elsid:
> Thank you for your investigation. Could you please send me the problematic test cases (3D models along with their corresponding parameters)? Also, could you specify where the Navmesh encountered issues?

It's not very simple to share the exact model. The problem appears in a scene composed out of many models and the contents is protected by the license not allowing to share it (it's not owned by me). Basically you need to buy the game. Although the program using the content to generate navmesh and find paths is open source. If you're interested I can share more details.

> I'm really sorry. I might not be able to guarantee solving this problem right now because my current schedule doesn't allow for in-depth investigation.

I'm interested if the change was intentional or expected. If it does not break anything, no need to worry.

Here is how it looks like:

recastnavigation_before

recastnavigation_after

elsid:
> Can you display a visual debugging mode similar to the "Walkable Voxels" in RecastDemo?

Not without writing code. I can try but this might take some time.

elsid:
Here are the screenshots with rendered walkable voxels (`duDebugDrawHeightfieldWalkable`) from different view point:

before:

recast_heightfield_before

after:

recast_heightfield_after

There is at least one voxel with no longer walkable surface now.

Recast/Source/RecastFilter.cpp
```diff
@@ -96,42 +96,47 @@ void rcFilterLedgeSpans(rcContext* context, const int walkableHeight, const int
...
-					for (neighborSpan = heightfield.spans[dx + dy * xSize]; neighborSpan; neighborSpan = neighborSpan->next)
+					for (neighborSpan = heightfield.spans[dx + dz * xSize]; neighborSpan; neighborSpan = neighborSpan->next)
 					{
-						neighborBot = (int)neighborSpan->smax;
+						int neighborBot = (int)neighborSpan->smax;
 						neighborTop = neighborSpan->next ? (int)neighborSpan->next->smin : MAX_HEIGHT;
 						
 						// Skip neighbour if the gap between the spans is too small.
-						if (rcMin(top, neighborTop) - rcMax(bot, neighborBot) > walkableHeight)
+						if (rcMin(top, neighborTop) - rcMax(bot, neighborBot) >= walkableHeight)
```
elsid, line 125: Why the comparison is changed from `>` to `>=`?

Recast/Source/RecastFilter.cpp
```diff
@@ -96,42 +96,47 @@ void rcFilterLedgeSpans(rcContext* context, const int walkableHeight, const int
...
 
 					// From minus infinity to the first span.
-					const rcSpan* neighborSpan = heightfield.spans[dx + dy * xSize];
-					int neighborBot = -walkableClimb;
+					const rcSpan* neighborSpan = heightfield.spans[dx + dz * xSize];
 					int neighborTop = neighborSpan ? (int)neighborSpan->smin : MAX_HEIGHT;
 					
 					// Skip neighbour if the gap between the spans is too small.
-					if (rcMin(top, neighborTop) - rcMax(bot, neighborBot) > walkableHeight)
+					if (rcMin(top, neighborTop) - bot >= walkableHeight)
```
elsid, line 112: Why the comparison is changed from `>`to `>=`?

Recast/Source/RecastFilter.cpp
grahamboree: If a voxel span is equal to the walkable height, it should allow the agent to traverse, no?

PR #684, opened 2024-01-09
[Detour] Compute tile pointers aligns correctly

m-bronnikov:
This commit fixes incorrect align computation for data in NavMesh's tile. Commit not changes user API.

Fixes: #669

### DIFF
--- a/Detour/Include/DetourCommon.h
+++ b/Detour/Include/DetourCommon.h
@@ -460,7 +460,26 @@ inline unsigned int dtIlog2(unsigned int v)
 	return r;
 }
 
+template <typename Type>
+int dtAlignOf()
+{
+    struct TestType 
+	{
+        Type t;
+        char c;
+    };
+
+    return static_cast<int>(sizeof(TestType) - sizeof(Type));
+};
+
 inline int dtAlign4(int x) { return (x+3) & ~3; }
+
+template <typename TypeToBeAlignedFor>
+int dtAlignForType(int x) 
+{ 
+	const int mask = dtAlignOf<TypeToBeAlignedFor>() - 1;
+	return (x+mask) & ~mask; 
+}
 
 inline int dtOppositeTile(int side) { return (side+4) & 0x7; }
 
--- a/Detour/Source/DetourNavMesh.cpp
+++ b/Detour/Source/DetourNavMesh.cpp
@@ -908,8 +908,12 @@ int dtNavMesh::queryPolygonsInTile(const dtMeshTile* tile, const float* qmin, co
 dtStatus dtNavMesh::addTile(unsigned char* data, int dataSize, int flags,
 							dtTileRef lastRef, dtTileRef* result)
 {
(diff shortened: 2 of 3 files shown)


m-bronnikov:
@grahamboree Hi, there is a fixed version of #670.

Could you please start the tests and take a look in case of success?

Detour/Include/DetourCommon.h
```diff
@@ -460,8 +460,27 @@ inline unsigned int dtIlog2(unsigned int v)
...
+int dtAlignOf()
+{
+    struct TestType 
+	{
+        Type t;
+        char c;
+    };
+
+    return static_cast<int>(sizeof(TestType) - sizeof(Type));
+};
```
m-bronnikov, line 473: Since `alignof` is not exist in c++98, I added simple implementation here

m-bronnikov:
> @grahamboree Hi, there is a fixed version of #670.
> 
> Could you please start the tests and take a look in case of success?

Looks like tests fail is not related to PR. Something gone wrong with `premake` deploy.

grahamboree:
Yeah looks like github had some issues.  Will try re-running them

grahamboree:
@m-bronnikov I re-ran the build jobs, but there are legitimate errors and warnings in the linux compilations.  I'm not sure why it's only showing up for premake, but it's possible that cmake is using different compiler flags than premake.

grahamboree:
Thanks for taking the time to submit this again.  After some careful consideration I'm not sure that this is really the best approach to address the issue you raised in #669

* This change modifies the tile buffer data layout even if you're _not_ using DT_POLYREF64.
* The solution here _technically_ avoids undefined behavior in un-aligned data access, but there wasn't unaligned access happening for cases that _weren't_ using DT_POLYREF64.
* By allowing the system to dictate the alignment of the data in the serialized tile buffer, it's creating a situation where loading the same data in the same application on two different hardware platforms (e.g. two consoles) can have different behavior if their alignment rules don't strictly match the machine the tile data was built on.  Platforms with matching alignment rules will behave correctly, but platforms that do not have matching alignment will read from the wrong places in the buffer and further cause undefined behavior if they read past the end of the buffer.
* A lot of this solution is predicated on a deep understanding of the undefined-behavior minutia in the C and C++ standards, which makes it difficult to work with safely.
* The real crux of the problem here is not that dtAlign4 is necessarily causing unaligned reads.  It's that it doesn't properly align things when using DT_POLYREF64.  

I think I'd much rather have either of the following solutions:
1.  Change `dtAlign4` to `dtAlign` and have it align to either 4 or 8 bytes depending on whether DT_POLYREF64 is set or not.  This is something I've specifically seen in embedded versions of Detour in commercial game engines.
2.  Change tile loading to copy data out of the buffer, rather than generate pointers into it.  This completely avoids any pointer alignment concerns.  The downside is that it uses twice the memory to add a tile to the navmesh.

PR #686, opened 2024-01-23
Fix unnecessary repetitions in the crowd

ikpil:
Maximum values should be checked after post-processing, as unnecessary repetitions may occur.


DetourCrowd/Source/DetourProximityGrid.cpp
```diff
@@ -163,9 +163,9 @@ int dtProximityGrid::queryItems(const float minx, const float miny,
 					// Item not found, add it.
 					if (i == end)
 					{
+						ids[n++] = item.id;
```
elsid, line 166: If `maxIds` is initially 0 and `ids` has 0 size (i.e. points to invalid memory) or `nullptr` this will lead to a crash due to `nullptr` dereference or out of bounds access.

PR #692, opened 2024-03-12
CMake: .pc: let GNUInstallDirs generate the paths

SomeoneSerge:
The concatenation approach results in paths like `libdir=${exec_prefix}//nix/store/...-recastnavigation-1.6.0/lib`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,9 +31,9 @@ install(FILES "${RecastNavigation_BINARY_DIR}/version.h" DESTINATION
 # Needed for recastnavigation.pc.in
 set(prefix ${CMAKE_INSTALL_PREFIX})
 set(exec_prefix "\${prefix}")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -31,9 +31,9 @@ install(FILES "${RecastNavigation_BINARY_DIR}/version.h" DESTINATION
 # Needed for recastnavigation.pc.in
 set(prefix ${CMAKE_INSTALL_PREFIX})
 set(exec_prefix "\${prefix}")
```
SomeoneSerge, line 33: What do you think about omitting `exec_prefix` from the `.pc.in` altogether?

grahamboree: I don't have strong opinions here but I'm all for simplifying the build pipeline

grahamboree:
LGTM ty!

PR #699, opened 2024-04-11
Add 16bit indices, cos angle precalculation and agent filtering options

TheBrooster:
* Added `rcMarkWalkableTriangles` and `rcClearUnwalkableTriangles` function overrides that allows the user to pass in 16bit index arrays.  Also added `..CosAngle` functions to allow for a precalculated cosine to be passed when called repeatedly.
* Added agent filter groups to allow an agent to ignore specific groups during neighbour collection.

### DIFF
--- a/DetourCrowd/Source/DetourCrowd.cpp
+++ b/DetourCrowd/Source/DetourCrowd.cpp
@@ -196,12 +196,17 @@ static int getNeighbours(const float* pos, const float height, const float range
 								pos[0]+range, pos[2]+range,
 								ids, MAX_NEIS);
 	
+	unsigned char ignoreGroupMask = skip ? skip->params.ignoreGroupFlags : 0u;
+
 	for (int i = 0; i < nids; ++i)
 	{
 		const dtCrowdAgent* ag = agents[ids[i]];
 		
 		if (ag == skip) continue;
 		
+		// Should `skip` ignore their neighbour `ag`?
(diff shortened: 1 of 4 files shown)


DetourCrowd/Source/DetourCrowd.cpp
```diff
@@ -196,12 +196,17 @@ static int getNeighbours(const float* pos, const float height, const float range
 								pos[0]+range, pos[2]+range,
 								ids, MAX_NEIS);
```
TheBrooster, line 197: Maybe we want to perform the agent filtering inside the `queryItems`?

DetourCrowd/Source/DetourCrowd.cpp
```diff
@@ -196,12 +196,17 @@ static int getNeighbours(const float* pos, const float height, const float range
...
+	unsigned char ignoreGroupMask = skip ? skip->params.ignoreGroupFlags : 0u;
+
 	for (int i = 0; i < nids; ++i)
 	{
 		const dtCrowdAgent* ag = agents[ids[i]];
 		
 		if (ag == skip) continue;
 		
+		// Should `skip` ignore their neighbour `ag`?
+		if ((ag->params.agentGroupFlags & ignoreGroupMask) != 0) continue;
```
TheBrooster, line 208: Maybe we want to perform the agent filtering inside the `dtProximityGrid::queryItems` call on line 195 above?

PR #718, opened 2024-06-18
Fix some compilation warnings

illwieckz:
Some warning fixes by a contributor of Unvanquished, initially pulled from:

- https://codeberg.org/bmorel/recastnavigation/commits/branch/main

I modified the first commit commit of that branch to:

1. extract a `chmod` change to a dedicated commit,
2. deleted a leftover,
3. move a change to the next commit it belongs to (partial downside squash).


illwieckz:
I'm reporting the macos and msvc errors reported by the CI to bmorel.

illwieckz:
> The remaining error happening with premake on Linux doesn't seem to be introduced by this branch:
> 
> ```
>  ../../Source/main.cpp:62:16: error: unused function 'createDebug' [-Werror,-Wunused-function]
> static Sample* createDebug() { return new Sample_Debug(); }
>                ^
> 1 error generated.
> ```

See this another PR for an attempt to fix that:

- https://github.com/recastnavigation/recastnavigation/pull/720

DetourTileCache/Include/DetourTileCacheBuilder.h
```diff
@@ -79,6 +79,13 @@ struct dtTileCachePolyMesh
 struct dtTileCacheAlloc
 {
 	virtual ~dtTileCacheAlloc();
+#if defined(__cplusplus) && __cplusplus >= 201103L
+	dtTileCacheAlloc() = default;
+	dtTileCacheAlloc( dtTileCacheAlloc const& ) = default;
+	dtTileCacheAlloc& operator=( dtTileCacheAlloc const& ) = default;
+	dtTileCacheAlloc( dtTileCacheAlloc && ) = default;
+	dtTileCacheAlloc& operator=( dtTileCacheAlloc && ) = default;
+#endif
```
grahamboree, line 88: Is this strictly necessary?
If it is, is there a version of this that would work for C++98 as well?

Detour/Include/DetourModernCpp.h
```diff
@@ -0,0 +1,24 @@
...
+#define DT_NULL NULL
+#define DT_OVERRIDE
+#endif
+
+template <typename TO, typename FROM>
+inline TO dtCast( FROM val )
+{
+	dtAssert( std::numeric_limits<TO>::min() <= val && val <= std::numeric_limits<TO>::max() );
+	return static_cast<TO>( val );
+}
```
grahamboree, line 22: This doesn't seem only applicable to "modern C++", so I'd be cool moving this to a `DetourCommon.h`

Also, I think a name like `dtSafeNumericCast` or `dtCheckedNumericCast` would be more appropriate here since this only works for numeric types, and it's checking that the value can be casted to the new type without losing information.

Recast/Include/RecastModernCpp.h
```diff
@@ -0,0 +1,24 @@
...
+#define RC_NULL NULL
+#define RC_OVERRIDE
+#endif
+
+template <typename TO, typename FROM>
+inline TO rcCast( FROM val )
+{
+	rcAssert( std::numeric_limits<TO>::min() <= val && val <= std::numeric_limits<TO>::max() );
+	return static_cast<TO>( val );
+}
```
grahamboree, line 22: Same thoughts here as the Detour version...

grahamboree (changes requested):
Thank you!  Love this change!  Especially using `RC_NULL` rather than `0` makes the code a lot clearer.  There's a few small things I'd love to tweak a bit before merging it.

PR #720, opened 2024-06-20
Fix premake linux demo build

illwieckz:
This is an attempt to fix this build error:

```
 ../../Source/main.cpp:62:16: error: unused function 'createDebug' [-Werror,-Wunused-function]
static Sample* createDebug() { return new Sample_Debug(); }
               ^
1 error generated.
```

I'm not sure this is correct. That line was commented out in:

- https://github.com/recastnavigation/recastnavigation/commit/f44e1515562ace8134241e66988773bc8ef503dc

Then it was removed later.

An alternative would be to entirely remove the `Sample_Debug` code.


grahamboree:
> An alternative would be to entirely remove the Sample_Debug code.

Yeah that sample looks like it's just some old scratch space for testing things.  I agree it should be removed.  I can let you do the honors if you like 😜

PR #723, opened 2024-07-10
[SAS5-2159] Defend the Drone GoTo Behaviour changes

albertoNK:
### JIRA Issue ID

This completes [SAS5-2159]

### What is the new behaviour

Drone should rise above zombie heads while traveling, and be completely ignored by zombies during this tim. When it Holds at a location to scan, zombies target it as a priority, as they do now.

The GoToDrone behaviour is what should be changed. Hold is fine right now.

### Other notes

added new crowd agent altitude and configuration parameters to let drones flyover

[SAS5-2159]: https://ninjakiwi.atlassian.net/browse/SAS5-2159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


grahamboree:
Hi! Thanks for this, but I'm not sure this is the right place for a change like this.  You're referencing some other Jira ticket that seems like it's for a project you're working on?  

If you have a more targeted fix/feature PR I'd gladly review it, but this seems very specific to your use case and not something I'd want to upstream into Recast without some discussion about what you're trying to accomplish here and whether that's generally applicable.

albertoNK:
yes, sorry, my fault :(

PR #728, opened 2024-08-11
Support enabling and disabling asserts for CMake builds

elsid:
Before this change asserts are enabled by default for CMake builds. This impacts performance. Add option to enable or disable asserts with default value depending on the build config. By default enable for Debug builds and disable for everything else.


grahamboree:
LGTM.  TY!

PR #734, opened 2024-10-10
Fix an issue where concave polygons may be generated during the rcBuildPolyMesh stage.

870751720:
Hello, after the mergeRegionHoles phase, we produced a contour as shown in the figure below. 
image
Points 23, 24, and 25 form a hole. In the case of the hole depicted above, we will produce a concave polygon during the rcBuildPolyMesh stage, which is not as expected.

I will provide a series of images demonstrating the results produced during each iteration of the triangulate function, where each image represents the shape of the remaining contour at that moment, along with the convexity status of each point. Large red dots indicate convex points, while small red dots indicate concave points.
polygon_0
polygon_1
polygon_2
polygon_3
polygon_4
polygon_5
polygon_6
polygon_7
polygon_8
polygon_9
polygon_10

It can be observed that when we reach Figure 6, point 25 is incorrectly judged as a convex point, which does not satisfy the definition of the ear clipping method. Subsequently, the convex information of point 25 persists into Figure 9, and then it is erroneously removed because it is a convex point that meets the removal criteria, as shown in Figure 10.

This ultimately leads to the generation of a concave polygon during the rcBuildPolyMesh stage.

The reason for this issue is that in Figure 5, when point 22 was removed, we only checked the convexity of points 20 and 23, without refreshing the convexity of point 25. The removal of point 22 caused a change in the convexity of point 25. To resolve this issue, I suggest that after each point removal, we perform a convexity check on all points to ensure that we do not mistakenly generate a concave polygon.

If you have better handling suggestions, I would appreciate discussing them with you. Thank you very much!


Recast/Source/RecastMesh.cpp
```diff
@@ -427,18 +427,16 @@ static int triangulate(int n, const int* verts, int* indices, int* tris)
...
-		if (diagonal(i, next(i1, n), n, verts, indices))
-			indices[i1] |= 0x80000000;
-		else
-			indices[i1] &= 0x0fffffff;
+		for (int i = 0; i < n; i++)
+		{
+			int i1 = next(i, n);
+			int i2 = next(i1, n);
+			if (diagonal(i, i2, n, verts, indices))
+				indices[i1] |= 0x8000;
```
Mauler125, line 436: Probably a small mixup with tilecache here and 2 lines down, shouldn't these feature the prevous constants (0x80000000 and 0x0fffffff)?

870751720: Yes, my mistake. Sorry, I subconsciously assumed that the values of the two were the same

870751720: I have submitted my modifications and thank you very much for pointing out my issues

PR #746, opened 2024-12-29
feat: add window resizing support

vbirds:
- Add SDL_WINDOW_RESIZABLE flag to enable window resizing
- Handle SDL_WINDOWEVENT_RESIZED event to update window dimensions
- Update OpenGL viewport and projection matrix when window is resized

This change allows users to resize the application window while maintaining proper aspect ratio and rendering.


grahamboree:
LGTM.  Thanks! 🎉

PR #756, opened 2025-01-20
`dtNavMesh::connectExtLinks` make max number of neighbors per edge configurable

Guokj-Git:
Background: In current version of `dtNavMesh::connectExtLinks`,  the method `findConnectingPolys` is called with the fixed parameter `maxcon` of 4. And all the neighbors per edge exceeding this limit will be dropped without any warnings.

What: I add a macro `DT_MAX_NEI_PER_EDGE` to support configurable number of neighbour per edge, and decide to raise an error when the number of neighbors exceeds this limit.

### DIFF
--- a/Detour/Include/DetourNavMesh.h
+++ b/Detour/Include/DetourNavMesh.h
@@ -56,6 +56,10 @@ typedef uint64_t dtTileRef;
 typedef unsigned int dtTileRef;
 #endif
 
+#ifndef DT_MAX_NEI_PER_EDGE
+#define DT_MAX_NEI_PER_EDGE 4
+#endif
+
 /// The maximum number of vertices per navigation polygon.
 /// @ingroup detour
 static const int DT_VERTS_PER_POLYGON = 6;
@@ -635,7 +639,7 @@ private:
 	void baseOffMeshLinks(dtMeshTile* tile);
 
 	/// Builds external polygon links for a tile.
-	void connectExtLinks(dtMeshTile* tile, dtMeshTile* target, int side);
+	dtStatus connectExtLinks(dtMeshTile* tile, dtMeshTile* target, int side);
 	/// Builds external polygon links for a tile.
 	void connectExtOffMeshLinks(dtMeshTile* tile, dtMeshTile* target, int side);
 	
--- a/Detour/Source/DetourNavMesh.cpp
+++ b/Detour/Source/DetourNavMesh.cpp
@@ -343,6 +343,9 @@ int dtNavMesh::findConnectingPolys(const float* va, const float* vb,
 				conarea[n*2+1] = dtMin(amax[0], bmax[0]);
 				con[n] = base | (dtPolyRef)i;
 				n++;
+			} else {
+				// No room for more connections.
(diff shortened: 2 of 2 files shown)


rob-myers:
Hey, is this related to off mesh connections? 

I've come across situations with too many such connections between two tiles. Some seem to be ignored.

However you are changing 'links' rather than off mesh links.

Guokj-Git:
> Hey, is this related to off mesh connections?
> 
> I've come across situations with too many such connections between two tiles. Some seem to be ignored.
> 
> However you are changing 'links' rather than off mesh links.

Hi, my current commit only focuses on external polygon links between tiles rather than off-mesh links.

After seeing your comment, I looked through the function `dtNavMesh::connectExtOffMeshLinks`, which should be a part of off-mesh links' buiding procedure, I found it skipping dealing with the situations when there are not enough memories to allocate a new off-mesh link.

``` cpp
    unsigned int idx = allocLink(target);
    if (idx != DT_NULL_LINK)
    {
        //
    }
    //should do something when allocation failed?
```

This skipping might leads to your situation, you can check it out.

Guokj-Git:
> Thanks for the idea, I'll have a look.
> 
> I'd also be interested in _why_ you want to increase max number of neighbors per edge.

I want to increase this number so that the method `findConnectingPolys` can record all neighbor polygons each edge shall have, thus help to make polygon-related queries on the navmesh works normally.

For example, if you called the method `dtNavMeshQuery::findStraightPath` on a navmesh with incomplete link record on some edges, the polygons corridor might come out with some weird detours.

rob-myers:
Thanks, that's interesting. Do you have a clear example?

image


For example
- the squares above amount to RecastDetour tiles
- the sub-triangles are the nav polygons.

How could the incomplete link record effect `findStraightPath` in the above example (if at all)?

Guokj-Git:
> Thanks, that's interesting. Do you have a clear example?
> 
> image
> For example
> 
> * the squares above amount to RecastDetour tiles
> * the sub-triangles are the nav polygons.
> 
> How could the incomplete link record effect `findStraightPath` in the above example (if at all)?

I draw an example for you:

On the left half of the picture, the `tileA` and `tileB` are adjacent and contain only few polygons. An example of a path found across these 2 tiles is also drew in blue lines.

After a few modifications on the `tileB`, e.g. adding obstacles, the number of polygons in `tileB` increases as illustrated on the right half of the picture. Unfortunately, the `edge_a` on the `tileA` only records a limited number of neighbor polygons `p0`, `p1` and `p2`, thus bringing a strange turning to the blue path found from `start` to `end`, which ought to be the same straight path on the left side of the picture.

IMG_20250221_172608

eugeneko:
Yes, I am having exact same problem.
But I don't fully agree with solution proposed in the PR.
I would rather have partially working navmesh + warning instead of unresolvable failure.
In my use case, _I don't know_ how many links I will have between two tiles. A few broken links are bad, but entire broken tile is so much worse.

Also, another problem is that we also need to make sure those links can be allocated, so maxLinks calculation should be updated.

PR #763, opened 2025-02-17
chore: ensure the bits of poly ref do not overflow.

sssooonnnggg:
If there is an overflow, the nav mesh query will return incorrect results.


grahamboree:
Thanks! 🎉

Issue #729, opened 2024-08-11
https://github.com/recastnavigation/recastnavigation/pull/672 claims to be refactor but changes behaviour


elsid:
See the discussion in the PR. Basically the library generates different navmesh before and after the change. It's not clear whether there are any correctness problems but changing the behaviour was not the author's intention. The problem is that this unexpectedly changes found paths and in my case breaks tests on different project. My suggestion is to revert the PR since the is no reaction on the provided info and the value of the PR itself is questionable.

elsid:
For the reference https://github.com/elsid/recastnavigation/commit/d5cd383970db8fa01d6117fdf985b92350f31b31 adds a test reproducing the problem. https://github.com/elsid/recastnavigation/actions/runs/14257052303 is a run before https://github.com/recastnavigation/recastnavigation/pull/672 and https://github.com/elsid/recastnavigation/actions/runs/14257055740 is for the run after. The test failed because of different number of found path points:

```
../../../Tests/Integration/Issue729.cpp:265: FAILED:
  REQUIRE( pathLen == 32 )
with expansion:
  25 == 32
```

elsid:
The change in [one comparison](https://github.com/recastnavigation/recastnavigation/pull/672/files#r2041116893) makes the difference. After additional refactoring it's now [here](https://github.com/recastnavigation/recastnavigation/blob/bd98d84c274ee06842bf51a4088ca82ac71f8c2d/Recast/Source/RecastFilter.cpp#L133). Why it was changed is unclear but changing it back fixes the issue:

https://github.com/elsid/recastnavigation/commit/903c322e623b465fea9c397d93833c7743e77af5

PR #772, opened 2025-04-13
Revert changes in comparison operators from PR #729

elsid:
https://github.com/recastnavigation/recastnavigation/pull/672 changed comparison operators from `>` to `>=` in 2 places for unclear reason which changed how navmesh is generated (https://github.com/recastnavigation/recastnavigation/issues/729).

Add test case affected by this change.


psi29a:
@grahamboree can you have a look please? :)

PR #786, opened 2025-07-17
Tiny linting on DetourNavMesh.cpp

astutejoe:
Noticed that when resolving issues on findConnectingPolys :)


grahamboree:
Thanks.  This kinda stuff is something I'm hoping to fix automatically with clangformat in the future.

PR #787, opened 2025-07-23
Added explicit dtNavMesh::teardown() function to allow in-place reuse…

sherief:
… of instances.

For cases where you want reuse and want to avoid dynamic allocation, this helps avoid workarounds such as using an `alignas()` buffer of char + placement new / explicit dtor.


grahamboree:
Thanks!  Is this `teardown` method materially different than calling the destructor explicitly?  Could we just add the two null pointer assignments to the existing destructor?

PR #791, opened 2025-08-06
use 8-bytes alignment when DT_POLYREF64 is defined

yantongl:
Our sanitized build caught a memory misalignment issue after I defined `DT_POLYREF64` to support larger world.

The problem is, 64 bit `dtPolyRef` changes `dtLink` to be 8-bytes alignment. However, in `addTile()` method `tile->links` is still 4-bytes aligned because memory is aligned using `dtAlign4()`. To suppress this issue, I replaced `dtAlign4()` by `dtAlign()` which uses 8-bytes alignment when `DT_POLYREF64` is defined.

As far as I can tell, this should not break runtime behavior and data compatibility. Looking forward to hear any feedback.


yantongl:
@Srekel  Thank you for pointing out the related issues. I saw issue #669 led to PR #684 , which was rejected in the end. My change actually matches the first idea @grahamboree mentioned in the end of #669, i.e. always use 8-bytes alignment when DT_POLYREF64 is defined. It removes the uncertainty #684 had. Hopefully this PR can lead to a fix to the problem.

I removed `dtAlign4` completely in this PR because I prefer to use the same alignment method everywhere in a build. It can be confusion if some code use 8-bytes alignment and other code use 4-bytes alignment.

PR #796, opened 2025-11-15
Fix crash on large-scale navmesh generation during dynamic scene changes

XYDXYD:
When building navigation meshes at large scale, the process may trigger an assertion and cause a crash. This issue becomes more pronounced in multiplayer or large-world scenarios where many players are actively modifying the environment at the same time, leading to oversized or frequently updated mesh data. This change improves stability by handling oversized mesh generation more safely and ensuring that dynamic scene updates do not cause unexpected failures.


grahamboree:
Thanks!  Can you please provide some more specifics about the issue this fixes, and/or an example repro case?

PR #806, opened 2026-01-29
Make RECASTNAVIGATION_ENABLE_ASSERTS do what it looks like it does

AnyOldName3:
There were some flaws with the existing implementation. Effectively, they meant that `RC_DISABLE_ASSERTS` was always defined unless someone explicitly set `RECASTNAVIGATION_ENABLE_ASSERTS` to one of CMake's truthy values.

* `option` only sets boolean options, so the `"$<IF:$<CONFIG:Debug>,ON,OFF>"` string was immediately converted to `OFF` as it's not literally *`1`, `ON`, `YES`, `TRUE`, `Y`, or a non-zero number*. Non-boolean options need to be declared with the `set` command in `CACHE` mode.
* Even if it *had* kept it as a string, it was fed to `if(NOT ...)`, which would have converted it to a boolean via the same rules. If you're using generator expressions, they can't be used for control flow, and must remain as strings until they're fed to a command that interprets them, and even then, that's deferred until after the `CMakeLists.txt` has fully executed and the configure phase is over and the generation phase has started.
* It's still slightly awkward as it needs negating in the place where it's used.

It's still got to be propagated to targets that consume recastnavigation as it's used in `RecastAlloc.h`, and there might be ODR violations if they use the templates that file defines with the preprocessor definition set differently. That can be a pain as if someone needs to toggle this to investigate an issue, it'll mean recompiling a whole game engine instead of just recast and the one or two translation units that include that header. It *might* be worth moving `rcVectorBase` and its derivatives to their own private header that's part of the build interface but not the install interface (and doing the same for the bits of the headers that define `dtAssert` and `rcAssert`), but I'm not doing that here as I'm just fixing something that already exists.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ option(RECASTNAVIGATION_TESTS "Build tests" ON)
 option(RECASTNAVIGATION_EXAMPLES "Build examples" ON)
 option(RECASTNAVIGATION_DT_POLYREF64 "Use 64bit polyrefs instead of 32bit for Detour" OFF)
 option(RECASTNAVIGATION_DT_VIRTUAL_QUERYFILTER "Use dynamic dispatch for dtQueryFilter in Detour to allow for custom filters" OFF)
-option(RECASTNAVIGATION_ENABLE_ASSERTS "Enable custom recastnavigation asserts" "$<IF:$<CONFIG:Debug>,ON,OFF>")
+set(RECASTNAVIGATION_ENABLE_ASSERTS "$<CONFIG:Debug>" CACHE STRING "Condition to enable custom recastnavigation asserts, evaluated as generator expression")
 option(RECASTNAVIGATION_ENABLE_FAST_MATH "Enable faster math calculations." OFF)
 
 # dll export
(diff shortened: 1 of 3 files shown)


AnyOldName3:
I can resolve the merge conflict myself if you let me know whether you'd prefer a rebase or a merge.

grahamboree:
Honestly I don't understand CMake nearly well enough to make informed choices here.  Kinda happy to just let the wisdom of the crowds dictate what should or should not be changed.  

Recast is _so darn simple_ to build I'm constantly astounded at how complex it is to get something that cmake is happy with.

Thanks again for the PR.  Rebasing would be preferable if it's not too much trouble.

AnyOldName3:
Usually, the CMake for a simple project should be simple, but smart people tend to go absolutely nuts and write the worst code they've ever written when they try and use it, and whenever you're using CMake, you'll have to interface with other people's CMake, so if they've written it badly, then yours ends up complicated, too. I've generally found https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/ is a good resource to send people. Recast's CMake is mostly better than average, but, for example, I expect that a lot of the SDL2-related commits that were merged recently are because you're using a CMake 2.x style `FindSDL2.cmake` module (which hasn't been updated for eight years) instead of just letting `find_package` find the CMake config files that SDL2 itself provides, and then trying to set up everything explicitly instead of just relying on `target_link_libraries` to do it automatically.

AnyOldName3:
As we've had someone comment on this and then delete their comment, I suspect this could catch someone else out, so I'll mention that one of the consequences of making `RECASTNAVIGATION_ENABLE_ASSERTS` go through `set` instead of `option` and therefore avoid CMake's string-to-bool conversion is that some of the literals that would previously have been mapped to true or false now are rejected as invalid generator expressions. If you want to make it always true or false, the direct way to do that is now to pass `1` or `0` as the value.

As it's taken as a parameter rather than computed, it might be sensible to pass the value through `$<BOOL:string>` to opt into CMake's string-to-bool conversion where it's used so there's no impact on any previously-working build scripts anyone might have.

PR #811, opened 2026-03-30
rcCalcBounds: fuse min/max loops into single pass

tommygrammar:
## Summary

Fuse the two-pass min/max loops in `rcCalcBounds` into a single pass that computes both bounds simultaneously. The change is semantically equivalent - output is identical to the original implementation.

## Motivation

`rcCalcBounds` currently iterates over the vertex array twice: once to compute minimums, once to compute maximums. A single fused pass produces the same result with fewer total instructions, fewer branches, and no change to cache behaviour or instruction-level parallelism.

## Changes

- `Recast/Source/Recast.cpp`: fused two-pass loop into single pass in `rcCalcBounds`
- `Tests/Recast/Tests_RecastBounds.cpp`: six new correctness tests for `rcCalcBounds`
- `Benchmark/rcCalcBounds_bench.cpp`: benchmark harness for timing comparison
- `Benchmark/CMakeLists.txt`: build configuration for benchmark executable
- `CMakeLists.txt`: added Benchmark subdirectory
- `Tests/CMakeLists.txt`: registered new Recast bounds test

## Behaviour Differences

None. Function signature, return type, and computed output are unchanged. This is a pure implementation optimisation.

## Public API Impact

None.

## Benchmark Results

Measured with `perf stat -r 20` at 100K iterations on Linux. Stable band taken from post-warmup runs.

| Metric | Original | Modified | Delta |
|---|---|---|---|
| ns/call (stable band) | 110–125 ns | 105–108 ns | ~8–12% reduction |
| Total instructions | 68,931,002 | 64,236,778 | −6.8% |
| Total cycles | 31,054,307 | 29,012,191 | −6.6% |
| IPC | 2.22 | 2.21 | unchanged |
| Branch count | 9,202,027 | 8,659,240 | −5.9% |
| Branch miss rate | 0.18% | 0.18% | unchanged |
| Cache miss rate | 39.77% | 39.13% | unchanged |

The instruction and cycle reduction is consistent with eliminating the redundant second loop pass. IPC and cache behaviour are unchanged, confirming the modification preserves the execution characteristics of the original.

## Tests

Six correctness tests added in `Tests/Recast/Tests_RecastBounds.cpp` covering:
- Basic bounding box computation
- Single vertex (bmin must equal bmax)
- Negative coordinates
- Mixed positive and negative coordinates
- All identical vertices (zero-volume bounds)
- Large vertex count (300 vertices with known min/max positions)

All six pass on both the original and modified build, confirming semantic equivalence.

**Existing test suite:** 25/26 cases pass on the modified build. The single failing case (`rcRasterizeTriangle` — degenerate triangles) reproduces identically on a pristine unmodified checkout — it is a pre-existing Catch2 runtime skip incompatibility with the build's exception configuration and is unrelated to `rcCalcBounds`.

### DIFF
--- a/Benchmark/CMakeLists.txt
+++ b/Benchmark/CMakeLists.txt
@@ -1,0 +1,18 @@
+
+add_executable(rcCalcBoundsBench
+    rcCalcBounds_bench.cpp
+)
+
+set_target_properties(rcCalcBoundsBench PROPERTIES
+    CXX_STANDARD 20
+    CXX_STANDARD_REQUIRED ON
+    CXX_EXTENSIONS OFF
+)
+
+target_include_directories(rcCalcBoundsBench PRIVATE
+    ${CMAKE_SOURCE_DIR}/Recast/Include
+)
+
+target_link_libraries(rcCalcBoundsBench PRIVATE
+    Recast
+)
--- a/Benchmark/rcCalcBounds_bench.cpp
+++ b/Benchmark/rcCalcBounds_bench.cpp
@@ -1,0 +1,50 @@
+#include <chrono>
+#include <cstdlib>
+#include <iostream>
+
+#include "Recast.h"
+
+int main()
+{
+    int iterations = 50000; //change the number of iterations here to observe different iteration performances
+    
+    float verts[] = {
+        1.0f, 2.0f, 3.0f,
+        0.0f, 2.0f, 5.0f,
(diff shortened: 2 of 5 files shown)


C3pa:
Looks like you didn't submit your changes to `Recast/Source/Recast.cpp`

